PR #3, opened 2020-05-14
Set correct edit urls

jamiebuilds:


davidmccabe:
Thanks for these fixes Jamie!

PR #9, opened 2020-05-14
docs: correct variable name in sample code

satoshicano:
~~isComplete -> isCompleted~~

isCompleted -> isComplete


satoshicano:
Right now, when I look under `docs/src/components /basic-tutorial/*`, I notice that "isComplete" is correct instead of "isCompleted". So I'll fix it now. Plz wait.

satoshicano:
The fix is complete. Please review

PR #15, opened 2020-05-14
Update getting-started.mdx

lcswillems:
I think it should `'textState'` instead of `textState`. Otherwise, I don't see where the variable `textState` is coming.

### DIFF
--- a/docs/docs/introduction/getting-started.mdx
+++ b/docs/docs/introduction/getting-started.mdx
@@ -77,7 +77,7 @@ function CharacterCounter() {
 }
 
 function TextInput() {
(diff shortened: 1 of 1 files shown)


lcswillems:
I am now thinking there is no typo... `textState` refers to this:

```
const textState = atom({
  key: 'textState', // unique ID (with respect to other atoms/selectors)
  default: '', // default value (aka initial value)
});
```

I got confused with `'textState'` string for key. I thought this was the way to refer to the atom.

drarmstr:
Thanks for trying out Recoil!

Issue #23, opened 2020-05-15
How to share state operation ?


xialvjun:
With ContextAPI :
```js
cosnt CounterContext = React.createContext(null);
const CounterProvider = (props) => {
  const [value, setValue] = useState(0);
  const add = useCallback(() => setValue(s => s+1));
  return <CounterContext.Provider value={{value, add}}>
    {props.children}
  </CounterContext.Provider>
}
```

But with recoil, can we :
```js
const counterDefault = {
  value: 0,
  add() {
    counterDefault.value += 1;
  },
}
atom({
  key: 'counter',
  default: counterDefault
})
```

Of course the code above is wrong, but how could we share state operation ? 
With context it's easy because operation closure has captured the real `setState` function. 
But with recoil, how to do it ?

By the way, @davidmccabe , can you open source your demo in the introduction video ? I still don't know how to use recoil to handle undefined number of state.

davidmccabe:
Here's how I would do it:

```
function useCounter() {
  const [value, setValue] = useRecoilState(countAtom);
  const add = useCallback(() => {setCount(x => x + 1)});
  return {value, add};
}
```

To make a family of atoms, just create a memoized function that returns a distinct atom for each state:

```
memoize(id => atom({
 key: `whatever${id}`,
 default: 123,
}));
```

where `memoize` is any memoization function you provide.

We are going to open-source a utility called `atomFamily`/`selectorFamily` that does this for you in the near future.

davidmccabe:
Actually I think it's already in the repo, the build just isn't configured to include it. If you can figure out how to set up the build to include the RecoilUtils module as a separate module e.g. `recoil/utils` in the built output, that'd be a nice PR.

tony-go:
I give a try https://github.com/facebookexperimental/Recoil/pull/33

Hope it help :)

xialvjun:
Thanks. 
In fact, I'm so dumb that I missed the point of your answer. I used to think we need to copy code in every component like:
```js
// counterAtom.js
export const counter = atom({
  key: 'counter',
  default: 0,
});
export const operations = (setValue) => ({
  inc: () => setValue(s => s+1),
  dec: () => setValue(s => s-1),
});

// myCom.jsx --- copy the code everywhere
const [value, setValue] = useRecoilState(countAtom);
const operations = useMemo(()=>counterOperations(setValue));
```

I forgot to use hook to extract logic to outside of component.
Just a mark.

PR #33, opened 2020-05-15
update rollup config and fix unresolved path

tony-go:
Hi 👋 

I followed this issue https://github.com/facebookexperimental/Recoil/issues/23, and I saw that I could help 😄 

I add a `/utils` in the bundle which contains `RecoilUtils.js`

I fix some unresolved paths (and a twice declaration in `recoil_values/Recoil_WaitFor.js` thanks to the rollup messages). I'm not a rollup expert so, let me know if I did something wrong.

_Question: There is no command to run tests?_

Thanks for this amazing job :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,6 @@ dist
 
 # Mac
 .DS_Store
+
+# Npm
+utils.js
--- a/package.json
+++ b/package.json
@@ -2,9 +2,10 @@
   "name": "recoil",
   "version": "0.0.7",
   "description": "Recoil - A React state management library for complex UIs",
-  "main": "dist/index.js",
+  "main": "dist/recoil.js",
   "files": [
-    "dist/recoil.js"
+    "dist/recoil.js",
+    "dist/utils/index.js"
   ],
   "repository": "https://github.com/facebookexperimental/recoil.git",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "name": "recoil",
   "version": "0.0.7",
   "description": "Recoil - A React state management library for complex UIs",
-  "main": "dist/recoil.js",
+  "main": "dist/index.js",
   "files": [
-    "dist/recoil.js",
+    "dist/index.js",
     "dist/utils/index.js"
   ],
   "repository": "https://github.com/facebookexperimental/recoil.git",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "Recoil - A React state management library for complex UIs",
   "main": "dist/index.js",
   "files": [
-    "dist/**/*.js"
+    "dist",
+    "utils.js"
   ],
   "repository": "https://github.com/facebookexperimental/recoil.git",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -4,8 +4,7 @@
   "description": "Recoil - A React state management library for complex UIs",
   "main": "dist/index.js",
   "files": [
-    "dist/index.js",
-    "dist/utils/index.js"
+    "dist/**/*.js"
   ],
   "repository": "https://github.com/facebookexperimental/recoil.git",
   "license": "MIT",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -35,7 +35,7 @@ const config = mode => [
   {
     input: 'src/Recoil.js',
     output: {
-      file: `dist/recoil.${mode}.js`,
+      file: `dist/Recoil.${mode}.js`,
       format: 'cjs',
       exports: 'named',
     },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,39 +3,55 @@ import babel from '@rollup/plugin-babel';
 import commonjs from '@rollup/plugin-commonjs';
 import {terser} from 'rollup-plugin-terser';
 
-const config = mode => ({
-  input: 'src/Recoil.js',
-  output: {
-    file: `dist/recoil.${mode}.js`,
-    format: 'cjs',
-    exports: 'named',
-  },
-  external: ['react', 'react-dom'],
-  plugins: [
-    babel({
-      presets: ['@babel/preset-react', '@babel/preset-flow'],
-      plugins: [
-        '@babel/plugin-proposal-nullish-coalescing-operator',
-        '@babel/plugin-proposal-optional-chaining',
-        '@babel/plugin-proposal-class-properties',
-      ],
-      babelHelpers: 'bundled',
-    }),
-    {
-      resolveId: source => {
-        if (source === 'React') {
-          return {id: 'react', external: true};
-        }
-        if (source === 'ReactDOM') {
-          return {id: 'react-dom', external: true};
-        }
-        return null;
-      },
+const getPlugins = mode => [
+  babel({
+    presets: ['@babel/preset-react', '@babel/preset-flow'],
+    plugins: [
+      '@babel/plugin-proposal-nullish-coalescing-operator',
+      '@babel/plugin-proposal-optional-chaining',
+      '@babel/plugin-proposal-class-properties',
+    ],
+    babelHelpers: 'bundled',
+  }),
+  {
+    resolveId: source => {
+      if (source === 'React') {
+        return {id: 'react', external: true};
+      }
+      if (source === 'ReactDOM') {
+        return {id: 'react-dom', external: true};
+      }
+      return null;
     },
-    nodeResolve(),
-    commonjs(),
-    mode === 'development' ? undefined : terser({ mangle: false }),
-  ],
-});
+  },
+  nodeResolve(),
+  commonjs(),
+  mode === 'development' ? undefined : terser({mangle: false}),
+];
 
-export default [config('development'), config('production')];
+const external = ['react', 'react-dom'];
+
+const config = mode => [
+  {
+    input: 'src/Recoil.js',
+    output: {
+      file: `dist/recoil.${mode}.js`,
+      format: 'cjs',
+      exports: 'named',
+    },
+    external,
+    plugins: getPlugins(mode),
+  },
+  {
+    input: 'src/RecoilUtils.js',
+    output: {
+      file: 'dist/utils/index.js',
+      format: 'cjs',
+      exports: 'named',
+    },
+    external,
+    plugins: getPlugins(mode),
+  },
+];
+
+export default [...config('development'), ...config('production')];
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -45,7 +45,7 @@ const config = mode => [
   {
     input: 'src/RecoilUtils.js',
     output: {
-      file: 'dist/utils/index.js',
+      file: `dist/utils/RecoilUtils.${mode}.js`,
       format: 'cjs',
       exports: 'named',
     },
--- a/src/npm/root-utils.js
+++ b/src/npm/root-utils.js
@@ -1,0 +1,7 @@
+'use strict';
+
+// this file will be added to the root after build
+// he allows users to require utils package as:
+// import { atomFamily } from 'recoil/utils'
+
+module.exports = require('./dist/utils/index.js');
--- a/scripts/postbuild.js
+++ b/scripts/postbuild.js
@@ -1,3 +1,4 @@
 const fs = require('fs');
 
 fs.copyFileSync('./src/npm/index.js', './dist/index.js');
+fs.copyFileSync('./src/npm/utils/index.js', './dist/utils/index.js');
--- a/scripts/postbuild.js
+++ b/scripts/postbuild.js
@@ -2,3 +2,7 @@ const fs = require('fs');
 
 fs.copyFileSync('./src/npm/index.js', './dist/index.js');
 fs.copyFileSync('./src/npm/utils/index.js', './dist/utils/index.js');
+
+// this line copy the file responsible for utils pretty import
+// require('recoil/utils')
+fs.copyFileSync('./src/npm/root-utils.js', './utils.js');
--- a/src/npm/index.js
+++ b/src/npm/index.js
@@ -1,7 +1,7 @@
 'use strict';
 
 if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./recoil.production.js');
+  module.exports = require('./Recoil.production.js');
 } else {
-  module.exports = require('./recoil.development.js');
+  module.exports = require('./Recoil.development.js');
 }
--- a/src/npm/utils/index.js
+++ b/src/npm/utils/index.js
@@ -1,0 +1,7 @@
+'use strict';
+
+if (process.env.NODE_ENV === 'production') {
+  module.exports = require('./RecoilUtils.production.js');
+} else {
+  module.exports = require('./RecoilUtils.development.js');
+}


davidmccabe:
Thanks!

We are still figuring out the OSS build system, etc.. If you want please feel free to add a command for running jest.

tony-go:
Hi :) I'm on it.

I got an issue when I run the test command with this line :

`const {act} = require('ReactTestUtils');`

I got a `Cannot find module 'ReactTestUtils' from 'src/recoil_values/__tests__/Recoil_atomWithFallback-test.js'`

I found some similarities with this https://fr.reactjs.org/docs/test-utils.html

How do you keep `'ReactTestUtils'` when you're running tests on your computer?

pmsandhu:
@tony-go You need to add `dist/utils/index.js` to the [files](https://docs.npmjs.com/files/package.json#files) field array in the `package.json` for it to be included in the npm package when published.

tony-go:
Hi @pmsandhu 

> @tony-go You need to add `dist/utils/index.js` to the [files](https://docs.npmjs.com/files/package.json#files) field array in the `package.json` for it to be included in the npm package when published.

Thanks :)

tony-go:
Hi @davidmccabe 

It seems that you probably had a special config to run your test in your local environment. Maybe could you share it? 

I got some errors like:
`Cannot find module 'Recoil_cacheWithReferenceEquality' from 'src/caches/__tests__/Recoil_cacheWithReferenceEquality-test.js'`
For line : `const cacheWithReferenceEquality = require('Recoil_cacheWithReferenceEquality')`

IMO it's probably just a config for resolve module in a specific way. As all imports wasn't handled properly any test failed.

Let me know :)

davidmccabe:
Hey @tony-go, let me know if you want help getting this merged.

tony-go:
Hi,

Yes, can you share me the way you run test on your computer?

For example babel doesn't resolve `require('Recoil_cacheWithReferenceEquality')` as rollup do with his plugin `@rollup/plugin-node-resolve`

recoil

I just want to be sure that anyone can run test :) me too ahahah

jonthomp:
@tony-go I think this is something to do with how Facebook require modules internally so I presume they will need to be changed in a way that still works there.
So that some of this can be merged sooner, perhaps split this up into 2 PRs, this one for the rollup build and a new one for the changes to tests?

tony-go:
Hi @jonthomp :)

> I think this is something to do with how Facebook require modules internally so I presume they will need to be changed in a way that still works there too.

Yeah, It might be something like that. 
I revert all my changes related to jest (:s) and rebase my job to master.

@davidmccabe 

Thanks, Guys

Shmew:
Hi @tony-go, I could be wrong, but shouldn't the `package.json` be like this?

```json
...
  "main": "dist/index.js",
  "files": [
    "dist/index.js",
    "dist/utils/index.js"
],
...
```

with a `src/npm/utils/index.js` being:

```js
'use strict';

if (process.env.NODE_ENV === 'production') {
  module.exports = require('./RecoilUtils.production.js');
} else {
  module.exports = require('./RecoilUtils.development.js');
}
```

and the rollup.config.js being:

```js
...
  {
    input: 'src/RecoilUtils.js',
    output: {
      file: `dist/utils/RecoilUtils.${mode}.js`,
      format: 'cjs',
      exports: 'named',
    },
    external,
    plugins: getPlugins(mode),
  },
...
```

and then the `postbuild.js`:

```js
const fs = require('fs');

fs.copyFileSync('./src/npm/index.js', './dist/index.js');
fs.copyFileSync('./src/npm/utils/index.js', './dist/utils/index.js');

```

tony-go:
🙈 Thanks a lot :)

I confess that this `postBuild` way to do, seemed a bit confusing when I rebased master.

It seems clearer with you review now :)

I finally got this:

Capture d’écran 2020-05-19 à 23 13 30

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Recoil - A React state management library for complex UIs",
   "main": "dist/index.js",
   "files": [
-    "dist/recoil.js"
+    "dist/index.js",
```
dustinsoftware, line 7: There are other files in dist that should be exported; try ‘dist/**/*.js’

You can confirm the result with ‘npm pack’ and inspecting the generated tar ball

tony-go:
@dustinsoftware Thanks so much, that's exactly what it misses IMO: a way to check that all files will be really published. I think `npm pack` solve this.

I update package.json according to your review, and it seems to be good :)

Capture d’écran 2020-05-20 à 09 46 32

dustinsoftware:
Verified by pulling down the artifact for this build and running from Node `require('recoil/dist/utils')`

That's probably good enough for this pass, we can iterate on a nicer import in the future if we need to.

tony-go:
@dustinsoftware 

What if we replace : 

```javascript
fs.copyFileSync('./src/npm/utils/index.js', './dist/utils/index.js');
```

by this : 

```javascript
fs.copyFileSync('./src/npm/utils/index.js', './dist/utils.js');
```

as utils.js will be at the dist root, we should get:

```javascript
require('recoil/utils')
```

dustinsoftware:
I’m fairly sure node won’t resolve that, but you could try it.

jonthomp:
@tony-go would you be able to merge this today?

tony-go:
@jonthomp I don't have right to merge it :/

tony-go:
Hi @dustinsoftware :) How are you? 

I just tried something with rollupjs. Can you tell me if we get something like `require('recoil/utils')` with it?

I can revert it if it's not the case.

jaredpalmer:
Hey all, so I run into this a lot because of tsdx. NPM imports are relative to the position of `package.json` and not the `main` entry point. To get `recoil/utils` as the import you have 2 options. 

1. Copy `package.json` from the root into the `dist` as a postbuild step and publish from that directory, fix `files` accordingly.
2. Keep `package.json` where it is, change `files` to `["dist", "utils.js"]`, make/copy an entry file `./utils.js` as follows:

```js
// ./utils.js
'use strict';

if (process.env.NODE_ENV === 'production') {
  module.exports = require('./dist/utils.production.js');
} else {
  module.exports = require('./dist/utils.development.js');
}
``` 

You can add `utils.js` to `.gitignore`, or just leave it hard coded. 

Now imports will work like you want:

```js
import { RecoilRoot } from 'recoil'
import { atomFamily } from 'recoil/utils'
```

dustinsoftware:
Cool! Yeah that approach sounds great.

tony-go:
Thanks a lot, @jaredpalmer :)

I apply the second method with a utils.js like this:

```js
// ./utils.js
'use strict';

module.exports = require('./dist/utils.index.js');
```
As `'./dist/utils.index.js'` already make the prod/dev choice for us.

I let you see the last commit guys! @davidmccabe @dustinsoftware @jonthomp @jimmyandrade  🎉

.gitignore
```diff
@@ -109,3 +109,6 @@ dist
 
 # Mac
 .DS_Store
+
+# Npm
+utils.js
```
jimmyandrade, line 114: Just a thought: shouldn't we `.gitignore` the `/dist` directory too?

rollup.config.js
```diff
@@ -3,39 +3,55 @@ import babel from '@rollup/plugin-babel';
...
+  mode === 'development' ? undefined : terser({mangle: false}),
+];
+
+const external = ['react', 'react-dom'];
+
+const config = mode => [
+  {
+    input: 'src/Recoil.js',
+    output: {
+      file: `dist/recoil.${mode}.js`,
```
jonthomp, line 38: A tiny thing but perhaps for consistency with the input filename and `RecoilUtils`,  `recoil` should be `Recoil`?

.gitignore
tony-go: Already in the file :)

rollup.config.js
tony-go: You're right :) done!

drarmstr:
Thanks for working on this!  We're reviewing shortly.

.gitignore
jimmyandrade: Great! And, sorry, I forgot to look at the entire content :)

src/npm/root-utils.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+// this file will be added to the root after build
+// he allows users to require utils package as:
```
ohansemmanuel, line 4:
```suggestion
// This allows users to require utils package as:
```

jimmyandrade: Why are we using “he”?

Shmew:
@davidmccabe any chance we can get this merged/published today?

src/npm/root-utils.js
tony-go: Im just waiting other review to change my code. Sorry for that by the way. I use French for construct my sentences before write it in English ahhaah.

jimmyandrade: No problem, @tony-go. My native language is Portuguese - which is also a strongly gender-oriented language - and this is something that can happen 😄

mondaychen:
Thanks @tony-go for contributing! I working hard to fix the syncing mechanism between github and FB internal repo. Once it's done (hopefully today), we'll be able to merge this.

mondaychen:
Hi @tony-go I'm ready on my side. Can you fix conflicts?

mgmeyers:
Hey @tony-go and crew, I pulled your branch down as I was interested in playing around with `atomFamily`. In doing so I believe I have uncovered an issue with creating two separate bundles.

Both `Recoil.*.js` and `RecoilUtils.*.js` both have `src/core/Recoil_node` as a dependency, but they each bundle the dependency separately. This means they each instantiate their own `nodes` map:

https://github.com/facebookexperimental/Recoil/blob/5f4491a69b8d1e383e49ddae06a1d4f3ac14c8e2/src/core/Recoil_Node.js#L58

So, once bundled, `RecoilUtils.*.js` can't access the node map created by `Recoil.*.js` and this error is thrown: `Missing definition for RecoilValue: "isToolbarPositionLockedState__"id-3"__withFallback""`

ryansolid:
If @mgmeyers is saying is correct, which a cursory glance of the rollup config shows both builds have the same externals, this gets a bit more complicated. One way of solving this is if the libraries need to share a single global state then the utils library should only pull in internal dependencies that are exposed from the recoil core package, and mark as an external in rollup. Otherwise anything shared needs to go into a 3rd build that is referenced by the other two as external and also in the dist.

I sort of get the whole desire to not put this stuff in core. But on the other hand isn't it sort of necessary the second you nest anything and not reconcile the whole list every time? I mean other than building it all out yourself that is. I've been attempting to do the simplest of demos (a Todo List where you can propagate a single selected state from the parent without updating every component) and it seems like selectorFamily is made for this scenario.

jonthomp:
I'm having a similar experience to @ryansolid and imagine passing query parameters to selectors is going to be a common use-case, especially when it comes to working with asynchronous data queries just like it's mentioned in the guide

briandonahue:
Ah, so this is probably related to #142 I filed the other day.

> Hey @tony-go and crew, I pulled your branch down as I was interested in playing around with `atomFamily`. In doing so I believe I have uncovered an issue with creating two separate bundles.
> 
> Both `Recoil.*.js` and `RecoilUtils.*.js` both have `src/core/Recoil_node` as a dependency, but they each bundle the dependency separately. This means they each instantiate their own `nodes` map:
> 
> https://github.com/facebookexperimental/Recoil/blob/5f4491a69b8d1e383e49ddae06a1d4f3ac14c8e2/src/core/Recoil_Node.js#L58
> 
> So, once bundled, `RecoilUtils.*.js` can't access the node map created by `Recoil.*.js` and this error is thrown: `Missing definition for RecoilValue: "isToolbarPositionLockedState__"id-3"__withFallback""`

dustinsoftware:
If there is shared internal state that has to be synced between Recoil and RecoilUtils, then yeah pulling this into a single bundle makes sense as a simple solution...

Rollup has pretty good support for exporting tree-shakable ES module bundles so if bundle size is the concern here I think that can be addressed. Eg. if downstream apps don't use the utils, they won't be included in the final bundle. Right now only a commonJS bundle is exported which IIRC will always be the full bundle size when compiled by downstream apps.

tony-go:
Hi all 👋  :)

I was off yesterday, sorry for the wait.

I'll take a look at the day.

jonthomp:
Fyi while the bumps in this are worked out, I have a [branch here](https://github.com/jonthomp/Recoil/tree/output-esm-and-bundle-utils) outputting esm instead of cjs, with the utils in the core bundle and persisting the built dist directory in the repo

tony-go:
> Fyi while the bumps in this are worked out, I have a [branch here](https://github.com/jonthomp/Recoil/tree/output-esm-and-bundle-utils) outputting esm instead of cjs, with the utils in the core bundle and persisting the built dist directory in the repo

> If there is shared internal state that has to be synced between Recoil and RecoilUtils, then yeah pulling this into a single bundle makes sense as a simple solution...

So we should lost the `recoil/utils` path, no?

jaredpalmer:
For those reading, I forked and combined all core and utils into main entry file and published to `jared-recoil`, which also include TS types that include selectorFamily and atomFamily

My suggestion is do the same in master right now. The fact that the docs don’t currently match the current package is really really bad and likely frustrating the crap out of folks who are playing in codesandbox, etc.

Shmew:
@tony-go that appears to be the case, @drarmstr have you all decided if that's acceptable?

mondaychen:
Hi folks! Per discussion of Recoil team, we will move everything in `RecoilUtils` into index, so people have easier access to them.

This change will land tomorrow, or this weekends the latest. Then we will publish a new version soon.

cc @davidmccabe @drarmstr

tony-go:
I was thinking that will be a great idea @mondaychen :) But I wasn't sure that It was acceptable, and it's a bit hard to get feedback from core team here.

So should we close this PR? 😢

drarmstr:
> I was thinking that will be a great idea @mondaychen :) But I wasn't sure that It was acceptable, and it's a bit hard to get feedback from core team here.
> 
> So should we close this PR? 😢

Thank you so much for @tony-go for working on this while we're getting everything in order with syncing internal/external releases, &c.  As mentioned, for most of us this is our first open source effort trying to keep it in sync with an internal repository.  We were originally thinking to keep the core package really light-weight and emphasize that the utilities are just helpers which are built on top of the core API.  But, after all these issues decided to keep things simple.  The docs will still segregate them, but the helpers there are small and commonly needed enough that it should be ok to include in the main package for convenience and awareness for now.  That said, we may revisit this; we do have plans for some additional helpers that will be less commonly needed that we expect to try to put in a secondary package.  But, we really want to unblock people now.  Many many thanks to @mondaychen for taking care of these build/packaging issues!

tony-go:
hi @drarmstr :)

Thanks for the feedback 👍 

I'll follow the evolution and hope to help later.

👍

sgzapata:
Dear colleagues, I am a PhD candidate in software engineering. I ask for your cooperation in answering this survey:
https://forms.gle/SkoVELsuWSH5gCrJ8

IT IS NOT SPAM. **It will only take you 10 minutes.**

Excuse me, **thank you very much for your help**. Best regards. 

Sergio Zapata (szapata@iinfo.unsj.edu.ar)
https://www.researchgate.net/profile/Sergio-Zapata-2
Universidad Nacional de San Juan - Argentina

Issue #34, opened 2020-05-15
"window is not defined" - SSR support


gschultz49:
I received this error when attempting to retrieve the value from a selector. This occurs in a Next.js 9.4 project, which is inherently server side rendered. 

Looking through the source code it looks like there are a few references to the window object that aren't available in SSR. Are there any plans to support SSR? I tried to work around it by checking for the presence of the `window` object but because I'm trying to conditionally use state it breaks the [rules of hooks ](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) :(

```
ReferenceError: window is not defined
    at r (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:712)
    at C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:1276
    at e (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:1290)
    at C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:19654
    at i (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:19929)
    at Object.c [as get] (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:19966)
    at x (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:4554)
    at C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:6719
    at Object.replaceState (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:10416)
    at C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:6684
    at Object.trace (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:4312)
    at getRecoilValueAsLoadable (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:6652)
    at n (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:13390)
    at r (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:13682)
    at Object.getRecoilState (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:13775)
    at useRecoilState (C:\git\nextjs-workflows\node_modules\recoil\dist\recoil.js:1:14988)
```

Example consumer code:
```

export const projectNameState = selector({
  key: "projectName",
[...]

PR #53, opened 2020-05-16
Avoid window reference error in SSR

sbaudray:
Fixes https://github.com/facebookexperimental/Recoil/issues/34

### DIFF
--- a/src/util/Recoil_isNode.js
+++ b/src/util/Recoil_isNode.js
@@ -9,6 +9,10 @@
 'use strict';
 
 function isNode(object: mixed): boolean {
+  if (typeof window === 'undefined') {
+    return false;
+  }
+
   const doc =
     object != null ? (object: $FlowFixMe).ownerDocument ?? object : document;
   const defaultView = doc.defaultView ?? window;


andreastoermer:
I think the code is wrong.

```
if (typeof window === 'undefined') {
    return false;
  }
```

If window is undefined, then we are in Node and not in Browser. So the correct result must be:

```
if (window === 'undefined') {
    return true;
}
```

sbaudray:
`window === 'undefined'` throws a `ReferenceError` because `window` is not a valid reference. `typeof` returns `undefined` when the reference is not valid.

andreastoermer:
ok, base on your response, my meaning is the same:

the check must be:

```
if (typeof window === 'undefined') {
    return true;
}
```

in node window is undefined, in browser not. the function name is isNode, so window must be undefined, if you will response the true answer for a isNode check.

sbaudray:
`isNode` doesn't refer to nodejs but to DOM Nodes as commented here: https://github.com/facebookexperimental/Recoil/blob/c38012d92a292a4a867163cf7b14cfb3debddf01/src/util/Recoil_deepFreezeValue.js#L45

andreastoermer:
Ok, but then maybe it's better to rename to function to isDomNode.

vjeux:
Thanks!

PR #55, opened 2020-05-16
update example input component

pankajpatel:
use `defaultValue` instead of `value` for initial value


sbaudray:
`value` is the right attribute for a [controlled component](https://reactjs.org/docs/forms.html#controlled-components). 
`defaultValue` is meant to provide the initial value for an[ uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html#default-values) which is not the case there

PR #67, opened 2020-05-17
Add development build

### LINKED ISSUE #61: Shipping un-minified source
It looks like the currently published version of recoil contains minified source. When debugging issues in recoil itself, it's more challenging to trace the flow of execution, even when using the devtools pretty-printer.

Would it be possible to ship both a development version alongside the minified production version? It looks like the react package takes this strategy.

Screen Shot 2020-05-16 at 19 27 52
Screen Shot 2020-05-16 at 19 27 33

dustinsoftware:
This should work on Windows too.

Tested with `yarn link` on a local project.

Fixes #61

### DIFF
--- a/src/npm/index.js
+++ b/src/npm/index.js
@@ -1,0 +1,7 @@
+'use strict';
+
+if (process.env.NODE_ENV === 'production') {
+    module.exports = require('./recoil.production.js');
+} else {
+    module.exports = require('./recoil.development.js');
+}
--- a/package.json
+++ b/package.json
@@ -2,12 +2,12 @@
   "name": "recoil",
   "version": "0.0.7",
   "description": "Recoil - A React state management library for complex UIs",
-  "main": "dist/recoil.js",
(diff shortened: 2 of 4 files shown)


src/npm/index.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+if (process.env.NODE_ENV === 'production') {
+    module.exports = require('./recoil.production.js');
+} else {
+    module.exports = require('./recoil.development.js');
+}
```
vjeux, line 7: Is this a common pattern to do that? I’ve never seen that before

dustinsoftware: Not sure how common it is, I copied this straight from React and it works well enough for this use case.

https://github.com/facebook/react/blob/master/packages/react/npm/index.js

vjeux: Thanks, if React is doing it, it's probably safe :)

PR #69, opened 2020-05-17
Disable variable name mangling in terser

dustinsoftware:
Based off of #67 to avoid conflicts.

This fixes an issue that was breaking a fresh create-react-app from working, due to CRA's own minification step running only in production mode.

Demo repo is [here](https://github.com/dustinsoftware/recoil-playground/tree/094862e722ec6ed016d4b723061df77c7db65fb3), run `yarn build && npx serve build` to see the isuse.

Screen Shot 2020-05-17 at 01 25 22

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,12 @@
   "name": "recoil",
   "version": "0.0.7",
   "description": "Recoil - A React state management library for complex UIs",
-  "main": "dist/recoil.js",
+  "main": "dist/index.js",
   "files": ["dist/recoil.js"],
   "repository": "https://github.com/facebookexperimental/recoil.git",
   "license": "MIT",
   "scripts": {
-    "build": "rollup -c"
+    "build": "rollup -c && node scripts/postbuild.js"
   },
   "peerDependencies": {
     "react": "^16.13.1",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -37,7 +37,7 @@ const config = (mode) => ({
     },
     nodeResolve(),
     commonjs(),
-    mode === 'development' ? undefined : terser(),
+    mode === 'development' ? undefined : terser({ mangle: false }),
   ],
 });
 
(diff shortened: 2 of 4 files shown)


jacques-blom:
From your investigating, does it seem like it is the combination of terser's mangling and CRA's own minification that is causing the issue? This is happening with Expo as well, but is happening in dev and production modes.

Also, I've found that setting `compress.reduce_vars` to false, with mangle still enabled, also solves the issue. I'm not sure which is the better option, given that `reduce_vars` improves runtime performance to some degree and `mangle` improves bundle size to some degree.

I think it'd be useful to understand why exactly this is happening to see if there is a more surgical fix.

It seems that the issue is originating here: https://github.com/facebookexperimental/Recoil/blob/c38012d92a292a4a867163cf7b14cfb3debddf01/src/recoil_values/Recoil_selector_NEW.js#L401

Terser seems to discard the `{current: state}` state container object and reassign a single state variable multiple times which, along `TreeState` being readonly, seems to be causing the issue.

dustinsoftware:
Does Expo also run terser in development mode?

My guess this is an incompatibility in terser itself when processing pre-minified files; this PR aims to improve compatibility as much as possible with the existing ecosystem while still stripping out white space and any other unnecessary logic in the production build. We could open an issue in the terser repo to confirm. 

From a quick read of the terser docs, reduce_vars should only optimize spots where terser sees only one assignment. However if the variable name mangling is causing terser to miss the other assignments, that seems like the root cause of the bug here. Other optimizations may be relying on the same variable detection logic and make more incorrect transformations. 

So I’m currently thinking that disabling mangling entirely will yield the best long term results. 

https://terser.org/docs/api-reference#compress-options

davidmccabe:
Thanks!!!

We are pretty out of touch with the JS open-source build systems. If you have suggestions about what we should be doing differently please feel free to share them.

donkeedong:
When will this be released?

PR #71, opened 2020-05-17
[docs] Update description and add example for useRecoilValue()

eriknguyen:
**Overview:**
This is simply to add an example for `useSetRecoilState` hooks in [API Reference](https://recoiljs.org/docs/api-reference/core/useSetRecoilState/)

### DIFF
--- a/docs/docs/api-reference/core/useSetRecoilState.md
+++ b/docs/docs/api-reference/core/useSetRecoilState.md
@@ -9,7 +9,7 @@ Returns a setter function for updating the value of writeable Recoil state.
 
 - `state`: writeable Recoil state (an [`atom`](/docs/api-reference/core/atom) or a _writeable_ [`selector`](/docs/api-reference/core/selector))
 
-This is the recommended hook to use when a component intends to write to state without reading it.
+This is the recommended hook to use when a component intends to write to state without reading it. If a component used the `useRecoilState()` hook to get the setter, it would also subscribe to updates and re-render when the atom or selector updated. Using `useSetRecoilState()` allows a component to set the value without re-rendering when the value changes.
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


docs/docs/api-reference/core/useSetRecoilState.md
```diff
@@ -7,6 +7,37 @@ Returns a setter function for updating the value of writeable Recoil state.
...
 
-- `state`: writeable Recoil state.
+- `state`: writeable Recoil state (an [`atom`](/docs/api-reference/core/atom) or a _writeable_ [`selector`](/docs/api-reference/core/selector))
+
+This is the recommended hook to use when a component intends to write to state without reading it.
 
 ### Example
+
+```javascript
+import { atom, useSetRecoilState } from 'recoil';
```
drarmstr, line 17: To standardize with our Prettier config.  We're trying to automate that as part of the tests for code at least.
```
import {atom, useSetRecoilState} from 'recoil';
```

docs/docs/api-reference/core/useSetRecoilState.md
```diff
@@ -7,6 +7,37 @@ Returns a setter function for updating the value of writeable Recoil state.
 
 ---
 
-- `state`: writeable Recoil state.
+- `state`: writeable Recoil state (an [`atom`](/docs/api-reference/core/atom) or a _writeable_ [`selector`](/docs/api-reference/core/selector))
+
+This is the recommended hook to use when a component intends to write to state without reading it.
```
drarmstr, line 12: "This is the recommended hook to use when a component intends to write to state without reading it.  If a component used the `useRecoilState()` hook to get the setter, it would also subscribe to updates and re-render when the atom or selector updated.  Using `useSetRecoilState()` allows a component to set the value without re-rendering when the value changes."

eriknguyen: Cool! Thanks for the more detail description :)

docs/docs/api-reference/core/useSetRecoilState.md
eriknguyen: I see, yeah automation would be great!

eriknguyen:
> Thanks for helping!

Updated based on your sugggestions.

PR #81, opened 2020-05-17
feature(test): Adds jest as test script

claudiopro:
## Adds `jest` as test script to `package.json`

This PR adds a test script to package.json, adding a few bits and bobs to ensure jest can parse Flow typed JS and modules are correctly resolved according to the aliases:

FB module name | OSS module name
------------ | -------------
React | react
ReactDOM | react-dom
ReactTestUtils | react-dom/test-utils

The PR also adds the `install-peers` dev dependency described in Yvonnick Frin's [post on dev.to](https://dev.to/yvonnickfrin/how-to-handle-peer-dependencies-when-developing-modules-18fa).

[fb-only] If you're a Facebook employee, the full test output is captured for your convenience at P131376227.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,19 @@
+{
+  "presets": ["@babel/react"],
+  "plugins": [
+    [
+      "module-resolver",
+      {
+        "root": ["./src"],
+        "alias": {
+          "React": "react",
+          "ReactDOM": "react-dom",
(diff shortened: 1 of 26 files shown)


claudiopro:
I may change my mind entirely and reuse the Rollup config for Jest. I am seeing https://github.com/algolia/rollup-jest-boilerplate that looks really interesting.

Issue #74, opened 2020-05-17
Prettier config


jacques-blom:
What Prettier config do you use internally? It'd be great if we could embed it in the project as a `.prettierrc`, and perhaps run `prettier --check` on PRs to make sure contributors have run Prettier.

claudiopro:
That's a desirable addition. Would you like to propose a PR 😄 ?
This is what we use at Facebook:

```json
{
  "arrowParens": "avoid",
  "singleQuote": true,
  "trailingComma": "all",
  "bracketSpacing": false,
  "jsxBracketSameLine": true,
  "overrides": [
    {
      "files": ["*.js", "*.jsx"],
      "options": {
        "parser": "flow",
      }
    }
  ]
}
```

jacques-blom:
Thanks! On it 👍

PR #83, opened 2020-05-17
Prettier

jacques-blom:
Fixes https://github.com/facebookexperimental/Recoil/issues/74

Adds Prettier as a dev dependency, and adds a `format` script that will format all project files. 

This PR also includes a `yarn format` run for files that didn't conform to the Prettier config, and updates the contributing guidelines.

**A few questions for the maintainers:**
1. Do we want to include Markdown files?
2. Are we fine with a simple `format` script like this for now?
3. I've used https://github.com/azz/pretty-quick in some of my projects. It runs Prettier on your staged files on commit, so you never have to worry about running it manually if your IDE doesn't run it on save for some reason, or you don't have an IDE integration set up at all. Is this something you'd be interested in adding? The only downside is that `git commit` needs to wait for Prettier, so committing can slow down as the project grows. Happy to submit a PR!


src/lib/Recoil_recoverableViolation.js
```diff
@@ -17,8 +17,10 @@ function recoverableViolation( // @oss-only
   message: string, // @oss-only
   projectName: 'recoil', // @oss-only
   {error}: {|error?: Error|}, // @oss-only
-): null { // @oss-only
-  if (__DEV__) { // @oss-only
+): null {
+  // @oss-only
```
vjeux, line 21: This is going to be a problem unfortunately, for the sync script to work (it's not right now), we need to have `// @oss-only` on every single line.

If you can add `// prettier-ignore` at the beginning of the function that would be great

src/lib/Recoil_invariant.js
```diff
@@ -13,9 +13,11 @@
 
 // @fb-only: const invariant = require('invariant');
 
-function invariant(condition: boolean, message: string) { // @oss-only
-  if (!condition) { // @oss-only
-    throw new Error(message) ;// @oss-only
+function invariant(condition: boolean, message: string) {
```
vjeux, line 16: Same here

.prettierrc
```diff
@@ -0,0 +1,15 @@
+{
+  "arrowParens": "avoid",
+  "singleQuote": true,
+  "trailingComma": "all",
+  "bracketSpacing": false,
+  "jsxBracketSameLine": true,
+  "overrides": [
+    {
+      "files": ["*.js", "*.jsx"],
```
vjeux, line 9: We don't have jsx files.

vjeux:
Thanks for doing this!

1) Yeah, I think we should.
2) That's fine for now. I think that it would be good to have also a lint step to make sure that people don't commit mis-formatted code. But doesn't need to be in this PR.
3) Yeah, would be nice!

If you can just add `// prettier-ignore` before the two `@oss-only` functions and remove `.jsx` then we can merge!

jacques-blom:
Thanks, @vjeux! Changes pushed. I'll submit another PR for pretty-quick, and look into how hard it would be to add a rule to prevent `@oss-only` comments from moving to a new line.

davidmccabe:
Thanks!!!

PR #86, opened 2020-05-17
Fix more tests

vjeux:
This is the current output after this PR: https://gist.github.com/vjeux/0050ee599a30c12e922e181c73518f2e

Tests:       43 failed, 110 passed, 153 total

I have no idea how the tests are working on www right now...


drarmstr:
Curious, what's different about this testing environment that it doesn't like an array for a React.Node that `renderElements` in www is fine with?

drarmstr:
I see, looks like we're just missing the missing key warning in www.

PR #90, opened 2020-05-18
[docs] document persistence.

chiscoder:
Hello, I added some basic example and documentation about persistence because there is none right now and you have to deep into the code to understand how to do it.

### DIFF
--- a/docs/docs/guides/persistence.mdx
+++ b/docs/docs/guides/persistence.mdx
@@ -1,0 +1,58 @@
+---
+title: State Persistence
+sidebar_label: State Persistence
+---
+
+
+Recoils makes it easy to persist your state.
+
+In order to persist your state you need two simple steps:
+
+First observe the changes to your atoms using useTransactionObservation hook (note: This is in unstable state as of now),
+it will tell you about what changed and you can simply save these changes to the storage of your preference.
+This is a basic example implementation:
+
+```javascript
+useTransactionObservation_UNSTABLE(({ atomValues, atomInfo, modifiedAtoms, transactionMetadata }) => {
+        for (const modifiedAtom of modifiedAtoms) {
+            Storage.setItem(modifiedAtom, JSON.stringify({ value: atomValues.get(modifiedAtom) }));
+        }
+    })
+```
+
+*Storage* could be [*LocalStorage*](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), *[AsyncStorage](https://github.com/react-native-community/react-native-async-storage)* or whichever storage you like.
+
+* *atomValues*: Gives you information about the current state of your atoms.
+* *modifiedAtoms* Gives you a map containing the modified atoms.
+* *transactionMetaData* Gives you metadata about the current change.
+
+
+After you ensure that you're saving yout state to your storage, you need to recover it, it's easy using the prop *initializeState* on *RecoilRoot* component.
+It's a function that give you set method that allows you to update atoms using their key.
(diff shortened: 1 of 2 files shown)


docs/docs/guides/persistence.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: State Persistence
+sidebar_label: State Persistence
+---
+
+
+Recoils makes it easy to persist your state.
```
atanasster, line 7: Recoil

docs/docs/guides/persistence.mdx
```diff
@@ -0,0 +1,58 @@
...
+```
+
+*Storage* could be [*LocalStorage*](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), *[AsyncStorage](https://github.com/react-native-community/react-native-async-storage)* or whichever storage you like.
+
+* *atomValues*: Gives you information about the current state of your atoms.
+* *modifiedAtoms* Gives you a map containing the modified atoms.
+* *transactionMetaData* Gives you metadata about the current change.
+
+
+After you ensure that you're saving yout state to your storage, you need to recover it, it's easy using the prop *initializeState* on *RecoilRoot* component.
```
atanasster, line 30: ‘saving your state’

docs/docs/guides/persistence.mdx
```diff
@@ -0,0 +1,58 @@
...
+It's a function that give you set method that allows you to update atoms using their key.
```
atanasster, line 31: ‘a function that provides a set method to update atoms given a key’

cgomezmendez:
Updated the PR with the fixes

drarmstr:
Thank you so much @cgomezmendez for taking the initiative to dive into the code and figuring out how to do persistence.  I only hesitate to publish docs on this API right now since I'm actively working on cleaning up the API and would kinda like to avoid whiplash so soon; hence them being marked as _UNSTABLE.  But, maybe it's ok to get this up with some stronger disclaimers that this is how people can explore right now, though expect it to change in the near future.

cgomezmendez:
> Thank you so much @cgomezmendez for taking the initiative to dive into the code and figuring out how to do persistence. I only hesitate to publish docs on this API right now since I'm actively working on cleaning up the API and would kinda like to avoid whiplash so soon; hence them being marked as _UNSTABLE. But, maybe it's ok to get this up with some stronger disclaimers that this is how people can explore right now, though expect it to change in the near future.

What about this at the beginning

**Note that the useTransactionObservation API is NOT stable, so use at your own risk. They may change at any time (as we're experimenting with them)**

docs/docs/guides/persistence.mdx
```diff
@@ -0,0 +1,58 @@
...
+            const promises = keys.map((key) =>
+                Storage.getItem(key)
+            );
+            Promise.all(promises).then((values) => {
+                for (let i = 0; i < promises.length; i++) {
+                    const key = keys[i];
+                    const value = JSON.parse(values[i]).value;
+                    set({ key }, value);
+                }
+                setIsReady(true);
```
alexandrzavalii, line 48: what `setIsReady` is doing? should there be a useState?

chiscoder: Yes, it's a useState setter, should I remove it or add the useState before the method?

alexandrzavalii: I think its better to remove it,

chiscoder: Done.

PR #97, opened 2020-05-18
[docs] use full todoList for stats, not filteredTodoList.

iketown:
seems like we'd want the stats on the full todoList, not on the filtered list.
otherwise you click on "unfinished" and the stats say that you're 0% complete, then you click on "finished" and the stats say 100% complete.


alexandrzavalii:
yeah actually was confused by this in the tutorial

PR #107, opened 2020-05-19
Pack all JS files in dist folder

dustinsoftware:
This was an oversight from #67

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,7 @@
   "version": "0.0.7",
   "description": "Recoil - A React state management library for complex UIs",
   "main": "dist/index.js",
-  "files": [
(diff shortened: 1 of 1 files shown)


dustinsoftware:
#33 also addresses this; if that lands first this can be closed.

dustinsoftware:
FWIW there is a fair amount of discussion on #33 so on second thought, landing this sooner would unblock releasing the current `master` branch.  cc @vjeux

vjeux:
This makes sense

dustinsoftware:
#195 appears to be a duplicate PR, should we land one of them soon?

dustinsoftware:
This landed in master, closing.

PR #108, opened 2020-05-19
React native

### LINKED ISSUE #99: React Native Support
The repo currently uses ReactDOM in order to communicate with the DOM. 

When will we be able to use Recoil within React Native?

dustinsoftware:
Fixes #99 

Disclaimer: stayed up late hacking this out, there are likely a few bugs.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,8 +6,6 @@
       {
         "root": ["./src"],
         "alias": {
-          "React": "react",
-          "ReactDOM": "react-dom",
           "ReactTestUtils": "react-dom/test-utils"
         }
       }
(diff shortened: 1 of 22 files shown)


dustinsoftware:
Hmm I see that there is a duplicate effort now in #114 .....

I'm probably not going to work on this any further until a maintainer weighs in on what the preferred approach is.

davidmccabe:
Thanks so much for doing this!

I don't really have any expertise in this area but this seems like a good approach in that it outputs separate bundles and doesn't require any action on the user's part. (Is my understanding accurate?)

I think we have to keep the capitalized module names for FB's internal build system.

dustinsoftware:
Correct; this solution should require no configuration to work with either web or native projects. 

I can keep the PascalCase names if that keeps it working with FB internal tooling.

dustinsoftware:
might pick this up later if #114 doesn't get merged, whatever approach we land on should probably be coordinated with @mondaychen. Closing for now.

PR #95, opened 2020-05-18
feature(lint): Adds Alex linter for inconsiderate writing

claudiopro:
**Summary**

This PR, similarly to facebook/react-native-website#1337, adds [alexjs](https://alexjs.com/) linting to Recoil in order to identify and prevent insensitive and inconsiderate language from landing on the docs website.

- Adds Alex
- Configures Alex rules

**Test Plan**

- Run `yarn lint-docs`
- Review lint violations

**Follow up**
- Run Alex in CI
- Fix lint violations

PR #109, opened 2020-05-19
Adds lint-staged

jacques-blom:
Adds `lint-staged`, which will format all staged files using Prettier. This means you can't accidentally commit unformatted files. We can also add other linters to this (e.g. ESLint) later on.

I also removed `jsx` from the file pattern since the repo doesn't contain JSX files.

An example commit with `lint-staged`:

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "prepare": "install-peers",
     "build": "rollup -c && node scripts/postbuild.js",
     "test": "jest src/*",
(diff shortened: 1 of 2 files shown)


dustinsoftware:
Lint-staged is awesome, this pr looks good to me...

claudiopro:
Oh, TIL! I see this builds on husky. Looks legit to me, thanks for contributing it!

claudiopro:
We could also add Alex to ensure docs do not contain insensitive language (see #95)

PR #114, opened 2020-05-19
React Native bundle

jacques-blom:
Fixes https://github.com/facebookexperimental/Recoil/issues/99

**Additions:**
- Add the ability to set/get a batch function, which defaults to `ReactBatchedUpdates`.
- Add `ReactBatchedUpdates` files for Web and Native that are imported to set the batch function to the renderer-specific function.
- Add a `native` target to Rollup, which resolves `ReactBatchedUpdates`.

**Result:**
This gives us out of the box support for React DOM and React Native, with the library user not needing to make any changes.

**Notes:**
I haven't changed the use of `unstable_batchedUpdates` in the tests. (In fact I haven't been able to get the suite to pass on `master`, which I'll have another go at later, and file an issue if I'm still stuck.) But it'd be great if the maintainers or anyone with more contexts on the tests could weigh in here. It seems like we should use `unstable_batchedUpdates` from `react-test-renderer` here.

In future, we could add a separate entry point (e.g. `'recoil/alternate-renderers'`) that doesn't set a batch function and doesn't depend on `react-dom` or `react-native`. We should be able to do this really easily because the infrastructure for setting the batching function to anything is available (and already exported in this PR).

**Prior art for this PR:**
- React Redux: [Alternate renderers entry point](https://github.com/reduxjs/react-redux/blob/master/src/alternate-renderers.js) and [get/set batch functions](https://github.com/reduxjs/react-redux/blob/master/src/utils/batch.js)
- @dustinsoftware's [PR](https://github.com/facebookexperimental/Recoil/pull/108) with a similar implementation 🙌


package.json
```diff
@@ -37,6 +37,7 @@
     "install-peers-cli": "^2.2.0",
     "jest-cli": "^26.0.1",
     "prettier": "^2.0.5",
+    "react-native": "^0.62.2",
```
jacques-blom, line 40: I've added RN as a dev dependency. I'm not actually sure what the right approach is here, because `react-dom` and `react-native` are optional dependencies (one of them is required).

A good approach might be to remove `react-dom` from the peer deps, too.

rollup.config.js
```diff
@@ -29,13 +32,23 @@ const config = mode => ({
...
+      extensions: target === 'native' ? ['.native.js', '.js'] : undefined,
+    }),
     commonjs(),
-    mode === 'development' ? undefined : terser({ mangle: false }),
+    mode === 'development' ? undefined : terser({mangle: false}),
   ],
 });
 
-export default [config('development'), config('production')];
+export default [
```
jacques-blom, line 49: It could be nice to split out the production and development builds, since I think it's rate you'd want to build both at the same time. A watch option would be great for development, too. I can imagine a `yarn dev` command. Happy to do a PR if this is something people would like to see :)

rollup.config.js
```diff
@@ -29,13 +32,23 @@ const config = mode => ({
...
         }
+        if (source === 'ReactNative') {
+          return {id: 'react-native', external: true};
+        }
         return null;
       },
     },
-    nodeResolve(),
+    nodeResolve({
+      extensions: target === 'native' ? ['.native.js', '.js'] : undefined,
```
jacques-blom, line 42: This will resolve `.native.js` files over `.js` files with the same name. This is how we have Rollup to import `ReactBatchedUpdates.native.js` when building for RN.

src/npm/index.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./recoil.production.js');
+  module.exports = require('./recoil.production');
```
jacques-blom, line 4: Removed the explicit extension so that the React Native bundler (Metro) imports `recoil.{mode}.native.js`. If this is too terse, we can add an `index.native.js` entry file and a `react-native` key to the package.json.

src/Recoil.js
```diff
@@ -76,4 +80,5 @@ module.exports = {
 
   // Other functions
   isRecoilValue,
+  setBatch,
```
jacques-blom, line 83: The library user can call `setBatch` and pass a batching function exported by whatever React renderer they use. Due to the current setup (the lack of an `alternate-renderers` entrypoint), they'll still need `react-dom` as a dependency.

package.json
dustinsoftware: It should work fine to make `react-dom` as a devDependency for the test runner and leave out `react-native` entirely

rollup.config.js
dustinsoftware, line 42: This seems error prone, if a `.native.js` file is missing or has a typo, the browser version would be resolved instead. Was there an issue you discovered with the rollup config proposed in #108?

rollup.config.js
```diff
@@ -3,14 +3,17 @@ import babel from '@rollup/plugin-babel';
...
-    file: `dist/recoil.${mode}.js`,
+    file:
+      target === 'web'
+        ? `dist/recoil.${mode}.js`
+        : `dist/recoil.${mode}.${target}.js`,
     format: 'cjs',
     exports: 'named',
   },
-  external: ['react', 'react-dom'],
+  external: ['react', 'react-dom', 'react-native'],
```
dustinsoftware, line 16: `react-native` should probably not be resolved unless the target is `native`, likewise for `react-dom` and `browser`. (See the config in #108 for how to pull this off)

rollup.config.js
```diff
@@ -29,13 +32,23 @@ const config = mode => ({
         if (source === 'ReactDOM') {
           return {id: 'react-dom', external: true};
         }
+        if (source === 'ReactNative') {
```
dustinsoftware, line 35: I get that this is following prior art, but this seems like an artifact that we could remove unless there is a compelling reason to keep the PascalCase aliases.

src/Recoil.js
```diff
@@ -29,6 +29,10 @@ export type {
   RecoilValueReadOnly,
 } from './core/Recoil_RecoilValue';
 
+const {unstable_batchedUpdates} = require('./util/ReactBatchedUpdates');
+const {setBatch} = require('./util/Recoil_batch');
+setBatch(unstable_batchedUpdates);
```
dustinsoftware, line 34: I like that this is exported; future-proofs the library a bit. We should probably add a test for this once they are passing.

src/npm/index.js
dustinsoftware: This works; however resolution problems are harder to debug later if they go wrong :) For instance, this may resolve the browser version by mistake depending on how the react-native bundler is configured.

Can we stick with a more explicit approach as in #108? Or does that not work with the React Native bundler?

src/util/Recoil_batch.js
```diff
@@ -0,0 +1,16 @@
+// Default to a dummy "batch" implementation that just runs the callback
+function defaultNoopBatch(callback) {
+  callback();
```
dustinsoftware, line 3: Is there a case where we would actually want to invoke the callback right away outside of a test context? Thinking this may hide bugs where the react-dom/react-native batching wasn't used by mistake....

We could throw an error here instead to surface the problem and make users provide their own `callback => callback()` function

src/Recoil.js
```diff
@@ -29,6 +29,10 @@ export type {
   RecoilValueReadOnly,
 } from './core/Recoil_RecoilValue';
 
+const {unstable_batchedUpdates} = require('./util/ReactBatchedUpdates');
```
davidmccabe, line 32: Do I understand correctly that this module will be resolved differently depending on the build, thus giving us a different default and not requiring setBatch to be called on either platform?

davidmccabe:
I'd love it if someone with experience in this area could weigh in on whether this or #108 is better long-term. What I want:

1) Users don't have to think about this
2) Users' only includes code for the appropriate platform
3) Users don't have to wait on us to use other renderers

Hirbod:
Will this or approch #108 also work with Expo on iOS and Android? Because there is no .native.js in Expo even though it is react-native.

rollup.config.js
jacques-blom: This just ensures that the module isn't bundled with the library. It won't include the module if it isn't imported, so only including it when bundling for native won't change the output.

rollup.config.js
jacques-blom: Resolving because we need to preserve PascalCase imports

src/Recoil.js
jacques-blom: Agreed

src/util/Recoil_batch.js
jacques-blom: Good idea. Will add 👍

jacques-blom:
It will, @Hirbod. Expo uses the same bundler as a bare React Native app, so will also resolve the `.native.js` bundle.

jacques-blom:
> I'd love it if someone with experience in this area could weigh in on whether this or #108 is better long-term. What I want:
> 
> 1. Users don't have to think about this
> 2. Users' only includes code for the appropriate platform
> 3. Users don't have to wait on us to use other renderers

Hey @davidmccabe. Both approaches are very similar. The difference with this one is that you can, as a library user, set the batching function to anything you want. So custom renderers are supported here.

To answer your specific questions:
1. You'll only have to think about it if you use a renderer other than React DOM or RN.
2. I'm 90% sure the library user's bundler (they'll most likely use Webpack or Metro) will only bundle code for the necessary platform. I'll double check using Webpack and Metro.
3. See above, and the exported `setBatch` function.

src/Recoil.js
jacques-blom: Correct. You'll either get `unstable_batchedUpdates` from React DOM or from React Native here depending on the platform. Then we pass that platform-specific batching function to `setBatch`.

brentvatne:
@Hirbod - there is a .native.js extension in Expo, it uses the same platform extensions as bare React Native but also adds .expo.js to that.

stephanoparaskeva:
Is there any reason this hasn't seen a merge?

jacques-blom:
Howdy @davidmccabe. Mind having a look at my [reply above](https://github.com/facebookexperimental/Recoil/pull/114#issuecomment-631704266) when you have a chance, please? It'd be great to get RN support merged in soon.

jacques-blom:
(or if you're able to, @drarmstr) 🙂

brentvatne:
@jacques-blom - could you ci errors? it might make it easier to merge

jacques-blom:
Hey @brentvatne. It's a linting error caused by code from `master`. [This PR](https://github.com/facebookexperimental/Recoil/pull/392) fixes it.

hamedmam:
Hey guys. Any luck that we get React Native support anytime soon ? I am starting a new React Native project  and hoping to use Recoil for state management of it

jacques-blom:
Hey @drarmstr,

Around when do you think you or one of the other maintainers will have a gap to take another look at this PR?

Sorry for the mention again. I don't mean to put pressure on the maintainers. I understand this is OSS, and you may be working on this outside of your normal work. But a lot of RN users seem to be getting stuck, so I think this is a big source of frustration for a lot of people. And this is one of the oldest open PRs, so it'd be great to get it through if you're happy with the code.

Thanks!

drarmstr:
@jacques-blom - We're currently checking with the React Native team to help validate these PRs, thanks for your patience!

jacques-blom:
> @jacques-blom - We're currently checking with the React Native team to help validate these PRs, thanks for your patience!

Awesome. Thank you!

a-eid:
@drarmstr any updates on this ?

drarmstr:
> @drarmstr any updates on this ?

No, unfortunately no updates yet.

setupReactBatching.js
```diff
@@ -0,0 +1,4 @@
+const {setBatch} = require('./src/util/Recoil_batch');
+const {unstable_batchedUpdates} = require('react-dom');
+
+setBatch(unstable_batchedUpdates);
```
drarmstr, line 4: Is this only used for testing?  Could you add some comments how this interacts and works with the call to `setBatch` in `Recoil_index.js` to the `util/ReactBatchedUpdates` version?  Hmm, seems that this is added as a `setupFiles` in the `package.json`.  @mondaychen could you review the packaging part of this?  (Internally we'd have to move this file to scripts)

src/util/Recoil_batch.js
```diff
@@ -0,0 +1,15 @@
+function defaultBatch() {
```
drarmstr, line 1: File needs standard copyright headers

src/Recoil_index.js
```diff
@@ -34,6 +34,10 @@ export type {
   SelectorFamilyOptions,
 } from './recoil_values/Recoil_selectorFamily';
 
+const {unstable_batchedUpdates} = require('./util/ReactBatchedUpdates');
+const {setBatch} = require('./util/Recoil_batch');
+setBatch(unstable_batchedUpdates);
```
drarmstr, line 39: Code should be placed after the `require` blocks.  There should be a prettier rule for this.  Since this is adding actual logic to `Recoil_index.js`, please add comments to explain.  It looks like this is meant to be for production while `setupReactBatching.js` is for testing?  Could this be moved to `Recoil_batch.js`?

src/util/ReactBatchedUpdates.native.js
```diff
@@ -0,0 +1,2 @@
+const {unstable_batchedUpdates} = require('ReactNative');
```
drarmstr, line 1: File needs standard copyright headers.  Rename to `Recoil_ReactBatchedUpdates.native.js`.

src/util/ReactBatchedUpdates.js
```diff
@@ -0,0 +1,2 @@
+const {unstable_batchedUpdates} = require('ReactDOM');
```
drarmstr, line 1: File needs standard copyright headers.  Rename to `Recoil_ReactBatchedUpdates.js`.

src/util/Recoil_batch.js
```diff
@@ -0,0 +1,15 @@
...
+
+let batch = defaultBatch;
+
+// Allow injecting another batching function later
+const setBatch = newBatch => (batch = newBatch);
+
+const batchUpdates = callback => batch(callback);
+
+module.exports = {
+  setBatch,
```
drarmstr, line 13: Since this is setting up the batcher instead of an individual batch, how about `setBatcher()`?

drarmstr:
@jacques-blom - We're not getting much interest from React Native so far, so we discussed just going ahead and trying to review and merge this PR directly.  Can you rebase and add a unit test of some sort?  I added a few other review comments. 
 @mondaychen could you also review the build targets and packaging aspect of this?

a-eid:
@jacques-blom hey, are u still working on this PR ?

mondaychen:
When rebasing, please note that we are now using a plain config array instead of a programmatic way of generating config objects. I imagine the new way should be easier for you to put in native related items.

jacques-blom:
Thanks, @drarmstr. Will get on it this weekend 👍

rollup.config.js
```diff
@@ -74,6 +80,28 @@ const configs = [
...
+  {
+    input: inputFile,
+    output: {
+      file: `native/recoil.js`,
+      format: 'es',
+      exports: 'named',
+    },
+    external: [...externalLibs, 'react-native'],
+    plugins: commonPlugins.map(plugin => {
+      // Replace the default nodeResolve plugin
```
jacques-blom, line 93: I'm not a huge fan of the way this looks, but it works. Another option would be to turn commonPlugins into a function that takes in `nodeResolve` options.

src/core/Recoil_Node.js
```diff
@@ -92,14 +92,14 @@ function registerNode<T>(node: Node<T>): RecoilValue<T> {
     // TODO Need to figure out if there is a standard/open-source equivalent to see if hot module replacement is happening:
     // prettier-ignore
     // @fb-only: if (__DEV__) {
-      // @fb-only: const isAcceptingUpdate = require('__debug').isAcceptingUpdate;
-      // prettier-ignore
-      // @fb-only: if (typeof isAcceptingUpdate !== 'function' || !isAcceptingUpdate()) {
-        // @fb-only: expectationViolation(message, 'recoil');
-      // @fb-only: }
+    // @fb-only: const isAcceptingUpdate = require('__debug').isAcceptingUpdate;
```
jacques-blom, line 95: It seems that Prettier doesn't ignore whitespace before comments, so it is removing the whitespace here when it's called by `lint-staged` during `pre-commit`. Is this fine or will it cause a problem when using this code internally?

setupReactBatching.js
jacques-blom: Yeah, we needed `setupReactBatching.js` for testing because the default batching function was being set in `Recoil_index.js` and we were importing individual module files directly in tests, meaning the batching fn wasn't set. But now that that's happening in `Recoil_batching.js`, we don't need this file anymore. 🙌

src/util/Recoil_batch.js
jacques-blom: Done and renamed.

src/Recoil_index.js
jacques-blom: I've moved this to `Recoil_batching.js`. 👌

jacques-blom:
Rebased and comments resolved, @drarmstr. Left one question about Prettier's interaction with `@fb-only` code. And CI is still failing because of a test that's also failing on `master`.

jacques-blom:
I have also added unit tests to test the behavior of the batching functions, and a test that asserts that the default batcher function is the one from React DOM.

Although this does test the basic batcher logic, it doesn't test whether a RN user would get the RN batcher function. `react-redux` tests this by spinning up a Jest environment with the `react-native` preset. This works because they export their package as an ES module with separate files, and the selective import happens at **project** build time, not **package** build time.

So for Recoil, we're building a separate `native` package, by telling Recoil to rather import a `.native.js` file so that we include the `react-native` `batchedUpdates` import in the `native` package. But `react-redux` exports their package with a file structure analogous to the following:
```
.
├── index.js
├── batcher
│   ├── batcher.js
│   └── batcher.native.js
```

And because the `.native.js` import happens when the user's project is built, we can test that behavior to test whether the native bather is imported.

Check out the react-redux PR that added this test: https://github.com/reduxjs/react-redux/pull/1444

We could explore exporting the Recoil ES build as separate files to achieve the same thing, or we could find a way of testing the bundled code. The former feels like a better solution, since exporting multiple files has other benefits, too.

Unless anyone has any other ideas? Or, if we're happy to not have a test for this behavior for now that's fine, too.

src/core/Recoil_Node.js
drarmstr: We'll see if this gets converted to uncommented lines with `// @fb-only` at the end when importing to our repository or we need to do that manually.  But, either way, since that source of truth is uncommented, then prettier should fix it when we do the internal diff.  So, this shouldn't be an issue.  The OSS only lines can be for the opposite reason, but we think we have a workaround for those using `{}` blocks.

drarmstr:
Thanks @jacques-blom!  I'm on holiday this week but will try to take a look with @mondaychen next week.

jacques-blom:
> Thanks @jacques-blom! I'm on holiday this week but will try to take a look with @mondaychen next week.

Thanks. Have a good break!

rollup.config.js
mondaychen: what happens if we have `extensions: ['.native.js', '.js']` for other builds?

Muhammad-Iqbal:
any updates on this. ?

rollup.config.js
jacques-blom: Adding this will replace any file named `x.js` with its native counterpart (`x.native.js`) if that native counterpart exists. So we should only specify this extensions rule for when we want the bundle to include code meant for RN.

a-eid:
@jacques-blom it's probably worth noting that on RN `fast-refresh` a "duplicate atom key ..." fatal error is thrown.

pmclachlan:
> @jacques-blom it's probably worth noting that on RN `fast-refresh` a "duplicate atom key ..." fatal error is thrown.

On the bright side, it only happens (at least to me) if a file affecting the atoms is changed... just mucking with styles etc is fine.

Thanks for all your work on this, by the way -- really looking forward to the RN support!

mondaychen:
I'm trying to test the build but failed. I did the following:

1. checkout your branch
2. run `yarn build` in recoil
3. run `yarn link` to register
4. go to my react-native test project (a simple project created with expo-cli), run `yarn link recoil`
5. add `import {atom} from 'recoil'`, rebuild

This is the error message I got:

```
Failed building JavaScript bundle.
Unable to resolve "recoil" from "App.tsx"
```

I'm not an expert in react-native. Anything I missed or did in a wrong way?

a-eid:
@mondaychen there's no linking needed, 
`yarn add recoil#react-native` should work.

mondaychen:
@a-eid that would install latest recoil, but not the build from this PR

pmclachlan:
> I'm trying to test the build but failed. I did the following:
> 
> 1. checkout your branch
> 2. run `yarn build` in recoil
> 3. run `yarn link` to register
> 4. go to my react-native test project (a simple project created with expo-cli), run `yarn link recoil`
> 5. add `import {atom} from 'recoil'`, rebuild
> 
> This is the error message I got:
> 
> ```
> Failed building JavaScript bundle.
> Unable to resolve "recoil" from "App.tsx"
> ```
> 
> I'm not an expert in react-native. Anything I missed or did in a wrong way?

As far as I know, the react native loader doesn't deal with symlinks nicely, so you can't link the library in like this.

mondaychen:
Now I'm using `"recoil": "file:../recoil"` instead, and got this error:

```
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

This error is located at:
    in RecoilRoot (at App.js:41)
    in App (at renderApplication.js:45)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:106)
    in RCTView (at View.js:34)
    in View (at AppContainer.js:132)
    in AppContainer (at renderApplication.js:39)

```

pmclachlan:
> Now I'm using `"recoil": "file:../recoil"` instead, and got this error:
> 
> ```
> Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
> 
> This error is located at:
>     in RecoilRoot (at App.js:41)
>     in App (at renderApplication.js:45)
>     in RCTView (at View.js:34)
>     in View (at AppContainer.js:106)
>     in RCTView (at View.js:34)
>     in View (at AppContainer.js:132)
>     in AppContainer (at renderApplication.js:39)
> ```

Yeah - I get this error at the moment too.  I'm not sure what's wrong, though...  my working theory (well, let's call it "not very educated guess") is that the react version needed by my expo/react-native and what is needed by this recoil version doesn't match (the recoil is a couple points newer).  I'm not able to upgrade the react right now so I was going to wait a bit before proceeding...  although I understand this isn't super helpful (sorry).

src/util/Recoil_batcher.js
```diff
@@ -0,0 +1,45 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails oncall+recoil
+ * @flow strict
+ * @format
+ *
+ * This is a stub for some integration into FB internal stuff
+ */
+
+const {unstable_batchedUpdates} = require('ReactDOM');
```
mondaychen, line 14: I believe you meant this
```suggestion
const {unstable_batchedUpdates} = require('./Recoil_ReactBatchedUpdates');
```

mondaychen:
In package.json let's remove `react-dom` from `peerDependencies` and add `peerDependenciesMeta `  following this example: https://github.com/reduxjs/react-redux/blob/master/package.json#L42-L49

a-eid:
@jacques-blom hey, I was wondering if there's any progress on this ? 
@mondaychen any plans on merging this soon ?

mondaychen:
As long as I can make the native build work on my machine and the tests pass, I can merge this.

bryan-hunter:
would love this functionality as well if it's close to merge-able :)

Proyoyo:
Agreed. We're about to start a new React Native project and would like to use this if it's about to be merged soon.

cuongn265:
Same here. I just hold my project stuff to wait for this PR to be merged these days. Hoop this will be merged soon.

drarmstr:
At this point we're just waiting for someone to help finish this PR.  We don't really use Recoil with React Native internally and it isn't officially supported.  But, we're happy to work with the community to add fixes and get it working.  This PR hopefully helps add that.  But, @mondaychen ran into an error when trying to build it locally and we haven't heard back yet from @jacques-blom, the original author.  If someone else is interested in helping, then we can hopefully merge this.

rdy:
I saw the issues described by @mondaychen when I also tried to test a local version with yarn link. I think this was due to having multiple react versions from the linked package and my destination repo.

I was able to put this into an internal private npm repository and after publishing, I have verified that I was able to include a Recoil root and useRecoilState in a very basic test within a react-native environment. 

I can do some additional testing to see if there are any issues. I’ll post back my results.

haswalt:
We're interested in this change too. I have a semi complex example that I can put together to test this integration. I'll put something together tomorrow if it helps get this merged.

santinucera:
@jacques-blom do you have any news about this?

jacques-blom:
Hi, all.

Sorry for the silence from my side. Yeah, unfortunately locally linking the two projects is a bit of a mission. So I've published this branch to the GH package registry as `@jacques-blom/recoil` for testing.

[Here](https://github.com/jacques-blom/recoil-test-react-native) is a sample RN project with this package installed.

To install it in your own RN project, do the following:
1. Create an `.npmrc` file at the root of your project, and add the following contents:
    ```
    registry=https://registry.npmjs.org
    @jacques-blom:registry=https://npm.pkg.github.com
    ```
    
    This will route the `@jacques-blom/recoil` install through the GH package registry.

1. Install `@jacques-blom/recoil`
    ```
    npm install @jacques-blom/recoil
    ```
    or
    ```
    yarn add @jacques-blom/recoil
    ```
1. Import everything from `@jacques-blom/recoil` instead of from `recoil`
    E.g. `import {RecoilRoot} from '@jacques-blom/recoil'`

Just a note that I only recommend installing this package to test whether this PR works. I'd wait until this is merged into the official package before using it in a real project.

@mondaychen, would you mind giving this a go and letting me know whether all looks good to you? Thanks!

PS: I have also rebased and pushed this PR.

jacques-blom:
> @jacques-blom it's probably worth noting that on RN `fast-refresh` a "duplicate atom key ..." fatal error is thrown.

^^ I also noticed this happening.

This is because of hot reloading. The problem here is that `recoverableViolation` uses `console.error` to display this notice, which triggers a red box error in RN. I've changed `recoverableViolation` to call `console.warn` instead, which displays the warning like this:
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-09-15 at 20 03 32

Unfortunately, this can still be pretty annoying if a bunch of these pile up in the yellow box.

I'm not sure whether this has unintended consequences, though. Maybe we can change it to call `console.warn` in RN, and `console.error` in a browser.

Let me know your thoughts, @drarmstr, and @mondaychen.

jacques-blom:
Also, here's a video of my sample RN app, using Recoil atoms and selectors: https://youtu.be/v3Y4pRIQfOU

santinucera:
You are my hero @jacques-blom, all we need is @mondaychen to review it now, right?.

src/util/Recoil_batcher.js
jacques-blom: Fixed, thanks.

haswalt:
I've seen a few other libs with this same issue on RN with fast reload. Logging as warnings would ease development on RN. Also using a solution like https://github.com/facebookexperimental/Recoil/issues/378#issuecomment-648753629 would also remove this warning as the key would always be unique. But that might be hiding it rather than fixing it.

mondaychen:
Thanks @jacques-blom. I'm on this.

mondaychen:
Not sure what happened, but in your package (I tried 0.0.14 and 0.0.13), native/recoil.js has `import reactDom from 'react-dom';`, thus is not working for me.

mondaychen:
My local test went pretty smoothly so far (after manually fixing the react-dom issue)!

> In package.json let's remove `react-dom` from `peerDependencies` and add `peerDependenciesMeta ` following this example: https://github.com/reduxjs/react-redux/blob/master/package.json#L42-L49

This should still be a good idea.

I'll do some further testing and import this PR tomorrow morning :)

rdy:
I also applied the change fixing the react-dom reference, as noted by @mondaychen. Everything seems to be working for me, I haven’t merged the latest changes here yet, but I can try.

jacques-blom:
Thanks @mondaychen. I didn't hit the `react-dom` error because I had it in my RN project's dependencies. I removed it and then got the error.

The problem was that I didn't do a new publish of my package after I fixed [this](https://github.com/facebookexperimental/Recoil/pull/114#discussion_r474312173).

`@jacques-blom/recoil@0.0.16` includes this fix, and it works.

davidmccabe:
Thanks so much for putting in all the work on this!

Is there a way to tell when module replacement is occurring on React Native? That would allow us to suppress the duplicate key error message.

haswalt:
@davidmccabe this might be related: https://github.com/redux-saga/redux-saga/issues/1961 looks like redux-saga had the same issue

jacques-blom:
> Thanks so much for putting in all the work on this!
> 
> Is there a way to tell when module replacement is occurring on React Native? That would allow us to suppress the duplicate key error message.

Redux Saga seemed to have issues with sagas being registered multiple times, but I haven't experienced any issues related to Fast Refresh on RN, so a check should be good enough.

The only downside to this, as far as I understand it, is the user won't see the warning if they have Fast Refresh enabled and there are actually two atoms with the same key.

What do you think, @davidmccabe?

Also, I couldn't find a way to check whether or not Fast Refresh is enabled. Do you think you could reach out to the RN from your side? Or perhaps someone else in the thread has some context on this.

Unfortunately `module.hot` is always truthy in RN nowadays, so a `!!module.hot` check won't work.

mondaychen:
So I'm merging this, with a few changes I made on internal tool

- Make .native.js file content hidden in internal repo (because internal repo does not have react-native and build system was unhappy)
- fix some comments & lints
- reverted changes in `expectationViolation`, instead, use `console.warn` in Recoil_Node for now.

Let's keep looking for an elegant solution for the duplicated atom key issue triggered by hot reload.

drarmstr:
Would someone here be able to help with a regression test for React Native so it doesn't accidentally break in a future release?

jacques-blom:
Thanks so much for everyone's contributions to this! It's super awesome that we got it merged.

@drarmstr, I'll have another go at a regression test. 👍 

What is the release process for Recoil, @mondaychen? Will this be manually released with other changes? And do you know roughly when the release will happen? Thanks.

mondaychen:
@jacques-blom thank you for working on this! It's a big step for Recoil.
Yes, we currently do manual releases.
The next release we probably want to bump the version number to 0.1.0. We plan to use a couple weeks to get more tests, documentation, examples and the blog post ready for this milestone. Will keep you posted!

haswalt:
@mondaychen any chance we could get a "next" release on npm? Running nightlies maybe?

mondaychen:
@haswalt Good idea. I'll talk to team and keep you guys posted.

mondaychen:
Hi everyone! FYI we've published a "nightly build" branch for testing purposes. If you want to try out recoil with react native before we release next version, try install the `nightly` branch with the "install git branch" feature of npm/yarn:
```
npm install https://github.com/facebookexperimental/Recoil.git#nightly
```
Or
```
yarn add https://github.com/facebookexperimental/Recoil.git#nightly
```

a-eid:
@mondaychen hey, thank you for your work. 
I was wondering when do you think RN will be officially supported ?

YOEL311:
Thank you for your wonderful work
When will ReactNative support be perfect?

[We have been updated](https://stackoverflow.com/a/64303707/9161478)

krakzk:
Have been using the 'recoil-react-native: 0.0.8-patch1" for development. It seems that the repo has been discarded maybe. Can we safely port to the nightly update or will there be any breaking changes?

haswalt:
We've been using nightly for development on ReactNative for a little while now and haven't experienced any issues.

krakzk:
@haswalt  Have you used the library [`](https://www.npmjs.com/package/recoil-react-native)

haswalt:
@krakzk no beacuse that package is really old, none official and there is ReactNative support built into Recoil

mondaychen:
@krakzk @haswalt the latest version on npm (0.1.2) has react native support.

thegauravthakur:
@mondaychen which package are you talking about?

krakzk:
@thegauravthakur The recoil official package instead of any others. Add in package as "recoil" : "0.1.2"

PR #121, opened 2020-05-20
fix(flow): Add flow configuration file

### LINKED ISSUE #115: Should we create .flowconfig?
How do you type check your codebase? Now I can write literally everything and it won't be typechecked

leushkin:
closes #115 

Have I forgotten something in config file?

PS Also I wanted to change `build` script to also run flow before it runs rollup, but currently we have ~160 flow issues, so I didn't do it.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,42 @@
+[ignore]
+.*/__tests__.*
+.*/node_modules.*
+
+[include]
+
+[libs]
+
+[options]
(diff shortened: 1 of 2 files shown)


davidmccabe:
Thanks! 

Once we get repo syncing working again (hopefully today) we can see whether the flow errors are due to config differences or actual errors that have been introduced since the syncing stopped.

PR #133, opened 2020-05-21
Fix badge URL in README

Daniel15:
Show the status of the official build rather than @dustinsoftware's, and make it link to the actions page rather than just showing the badge image on click.


dustinsoftware:
Thanks for raining on my parade Daniel ;)

PR #150, opened 2020-05-23
Fix global variable __DEV__

mondaychen:


rollup.config.js
```diff
@@ -34,7 +35,8 @@ const config = mode => ({
     },
     nodeResolve(),
     commonjs(),
-    mode === 'development' ? undefined : terser({ mangle: false }),
+    replace({__DEV__: JSON.stringify(mode === 'development')}),
```
dustinsoftware, line 38: This seems to match the React rollup config, so looks good to me. Can we add a test verifying that DEV mode is active in Jest tests?

https://github.com/facebook/react/blob/master/scripts/rollup/build.js#L426

PR #151, opened 2020-05-23
[docs] Use splice to make the docs more intuitive

GunaShekar02:
The atoms section in the basic tutorial currently uses explicitly defined functions to replace and delete values from an array, which is exactly what the native "splice" method is used to do. 
Changing the tutorial to use splice instead reduced the number of lines of code and would also make it more intuitive for a Javascript developer to understand.

### DIFF
--- a/docs/docs/basic-tutorial/atoms.mdx
+++ b/docs/docs/basic-tutorial/atoms.mdx
@@ -92,16 +92,15 @@ function TodoItem({item}) {
   };
 
   const toggleItemCompletion = () => {
-    const newList = replaceItemAtIndex(todoList, index, {
-      ...item,
-      isComplete: !item.isComplete,
-    });
+    const newList = todoList;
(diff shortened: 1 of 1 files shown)


dustinsoftware:
I tested this out locally just to re-verify my understanding:

```
> let a = [1,2,3]
undefined
> a.splice(0,1)
[ 1 ]
> a
[ 2, 3 ]
```

Was just reminded of a surprising behavior: `splice` does _not_ return the mutated array; it returns the items removed. While this documentation PR doesn't make this mistake, this is a subtle error that downstream consumers of this library could make.

For this reason I think we should avoid using `splice` in the docs...

drarmstr:
Agree with @dustinsoftware that `splice()` isn't the most intuitive, and I think using helpers with explicit names like `replaceItemAtIndex()` and `removeItemAtIndex()` is more explicit on the intent of the code for documentation purposes.  It is also best to avoid mutation to keep things simple.  Closing this PR. cc @csantos42

PR #153, opened 2020-05-24
Fix call stack size exceeded when set atom value with object that has circular reference

### LINKED ISSUE #54: Infinity loop occurs if the value to be set has a circular reference
Hi there,

I've found infinity loop problem if the value to be set has a circular reference.
Is it bug or limitation of recoil?

Minimum reproduction:
https://codesandbox.io/s/intelligent-matsumoto-vsw7m?file=/src/App.js

## Actual behavior:

```
RangeError: Maximum call stack size exceeded
```

## Expected behavior:

The value can be set without error.

n3tr:
## Background

fix #54

Set atom value with an object that has circular reference inside will cause an error

```
RangeError: Maximum call stack size exceeded
```

## What's changed

- In `deepFreezeValue`  Check if a property is already frozen

### DIFF
--- a/src/recoil_values/__tests__/Recoil_atom-test.js
+++ b/src/recoil_values/__tests__/Recoil_atom-test.js
@@ -56,7 +56,7 @@ test('atom can store null and undefined', () => {
   expect(get(myAtom)).toBe('VALUE');
 });
 
-test('atom can store a circular reference object ', () => {
+test('atom can store a circular reference object', () => {
   class Circular {
     constructor() {
       this.self = this;
--- a/src/recoil_values/__tests__/Recoil_atom-test.js
+++ b/src/recoil_values/__tests__/Recoil_atom-test.js
@@ -55,3 +55,19 @@ test('atom can store null and undefined', () => {
   act(() => set(myAtom, 'VALUE'));
   expect(get(myAtom)).toBe('VALUE');
 });
+
+test('atom can store a circular reference object ', () => {
+  class Circular {
+    constructor() {
+      this.self = this;
+    }
+  }
+  const circular = new Circular();
+  const myAtom = atom<Circular>({
+    key: 'atom',
+    default: undefined,
+  });
+  expect(get(myAtom)).toBe(undefined);
+  act(() => set(myAtom, circular));
+  expect(get(myAtom)).toBe(circular);
+});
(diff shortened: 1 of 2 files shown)


dustinsoftware:
Cool! Seems like a pretty straight forward fix.

src/util/Recoil_deepFreezeValue.js
```diff
@@ -64,7 +64,7 @@ function deepFreezeValue(value: mixed) {
 
   Object.freeze(value); // Make all properties read-only
   for (const prop in value) {
-    if (value.hasOwnProperty(prop)) {
+    if (value.hasOwnProperty(prop) && !Object.isFrozen(prop)) {
```
drarmstr, line 67: This breaks `deepFreezeValue()` because it is checking if the property string key is frozen instead of the property itself.  strings are always frozen so it essentially turns off all freezing of properties.  It causes the tests in `Recoil_core-test.js` to fail.  We can easily fix this by looking up the property itself.

This also isn't perfect, since if we encounter a non-recursive, previously frozen, object that contains unfrozen objects, they won't get frozen.  But, probably a small corner case not worth the extra cost of keeping track of visited objects to otherwise avoid infinite recursion.

src/recoil_values/__tests__/Recoil_atom-test.js
```diff
@@ -55,3 +55,19 @@ test('atom can store null and undefined', () => {
   act(() => set(myAtom, 'VALUE'));
   expect(get(myAtom)).toBe('VALUE');
 });
+
+test('atom can store a circular reference object', () => {
```
drarmstr, line 59: Need to update the test to fix Flow errors, I can fix that when fixing the diff.  Let's see if our github sync scripts are working now....

drarmstr:
I've fixed the diff locally for our internal repository, but seems it doesn't update the GitHub PR.  Let's find out what happens when we approve and land it there if it then properly syncs with GitHub.

src/util/Recoil_deepFreezeValue.js
n3tr: You are right! My bad, It should be `Object.isFrozen(value[prop])` instead.

src/recoil_values/__tests__/Recoil_atom-test.js
n3tr: Interesting, Flow doesn't show any errors on my end. I guess it should be a missing property error on Circular class, there should be something wrong with my setup.

PR #154, opened 2020-05-24
Rename .babelrc to babel.config.js to fix test suite failed to run

n3tr:
## Background

> Not sure if this only happened only on my end

Unable to run test via `yarn test` due to Jest error parsing flow type:

```
SyntaxError: /Github/Recoil/src/adt/Recoil_ArrayKeyedMap.js: Unexpected token, expected "{" (24:19)

      22 | const emptyMap = new Map();
      23 |
    > 24 | class ArrayKeyedMap<V> {
   ...
```

This is probably because `babel-jest` doesn't pick up a babel configuration properly (`.babelrc`) due to the project has 2 `package.json` files (one in root and one in `/src`). Since the `.babelrc` is per package configuration which makes `babel-jest` won't transform any files inside the `/src` folder.

**There are 2 approaches to fix this:**

- Remove `package.json` from`/src` (Not sure if we do need it?)
- Rename `.babelrc` to `babel.config.js` which will act as a project-wide configuration (in case we don't want to remove `src/package.json`) **propose in this PR**

## What's changed

- Rename the babel root configuration from `.babelrc` to `babel.config.js` per recommendation in [jest documentation](https://jestjs.io/docs/en/getting-started#using-babel)

## Result

Test suite is able to run

```
Tests:       43 failed, 110 passed, 153 total
Snapshots:   0 total
Time:        4.378
```


dustinsoftware:
I don't think this is the right approach; the tests only started failing when https://github.com/facebookexperimental/Recoil/pull/148 was merged.

The cause seems to be a duplicate `package.json`, I have PR'd removing that here and can confirm locally that the issue is resolved

https://github.com/facebookexperimental/Recoil/pull/156

n3tr:
@dustinsoftware 

Agreed, Removing the duplicated `package.json` is better solution.

I just don't know the reason behind why we need 2 `package.json`, probably one get used internally at FB. So I proposed this solution instead of removing it.

Actually merge both PRs would be fine as well.

mondaychen:
The explanation on why we had it: https://github.com/facebookexperimental/Recoil/pull/156#issuecomment-633302768

I'm working on the internal config, and this will be fixed once #196 is merged. Thanks @n3tr and @dustinsoftware for contribution!

mondaychen:

PR #156, opened 2020-05-24
Remove duplicate package.json

dustinsoftware:
This also seems to fix Jest not resolving the babelrc correctly.


mondaychen:
This file is necessary for Facebook's internal repo. We can't just remove it. If this would break jest and there's no alternative fix, I can try to remove it by blacklisting it in some config file.

dustinsoftware:
Yeah I think blacklisting it would be good with whatever sync tool is in use... as an outsider it's confusing which `package.json` file should be trusted as the source of truth for the version shipped to `npm`.

mondaychen:
this will be done by #196

mondaychen:

PR #159, opened 2020-05-25
[docs] initial docs for syncing state

### LINKED ISSUE #143: Can you share an example of useSubscription. How to subscribe and unsubscribe using selector

### LINKED ISSUE #149: How to set a new state using async API call
I'm trying to implement basic features to a project using Recoil, just to test the library. I'm really liking what I'm seeing. However, I have a doubt regarding to updating a state based on an async API call, i.e., I created the following `selectors` to fetch from API:

```
export const teamsState = selector({
    key: "teams",
    get: async ({ get }) => {
        try {
            const response = await api.get("/teams");
            return response.data.teams;
        } catch (e) {
            throw e;
        }
    },
});

export const teamById = (id) =>
    selector({
        key: "teamById",
        get: ({ get }) => {
            const teams = get(teamsState);

            return teams.filter((team) => team._id === id)[0];
        },
    });

```

The `teamsState` fetch a list of `teams` from the API, and the `teamById` returns the selected `team` from the list of teams.

My doubt comes now. If I need to add a new team, how should I proceed?

I have a route to add a team in my backend, and the response is the object team I added, not a list of teams. So, how should I update the `teamsState` to reflect this added object? Should I get all teams again, making an API call to get the list of teams (this doesn't seem the right choice, but would work)?

Great work on bringing this new library to the React environment!

drarmstr:
resolves #143 
resolves #149

### DIFF
--- a/docs/docs/guides/asynchronous-state-sync.md
+++ b/docs/docs/guides/asynchronous-state-sync.md
@@ -2,3 +2,185 @@
 title: Asynchronous State Sync
 sidebar_label: Asynchronous State Sync
 ---
+
+Recoil [atoms](/docs/api-reference/core/atom) represent local application state.  Your application may have remote or server-side state as well, such as via a RESTful API.  Consider synchronizing the remote state with Recoil atoms.  Doing this allows you to easily access or write to the  state from React components using the `useRecoilState()` hook, or use that state as input to the Recoil data-flow graph for other derived state selectors.  If you're looking to [query a database or server for read-only data](asynchronous-data-queries), consider using asynchronous selectors.
+
+## Local State Example
+
+This example provides the friend status as local state only.
+
+```jsx
+const currentUserIDState = atom({
+  key: 'CurrentUserID',
+  default: null,
+});
+
+function CurrentUserInfo() {
+  const [currentUserID] = useRecoilState(currentUserIDState);
+  return <div>Current User: {currentUserID}</div>;
+}
+```
+
+## Sync State From Server
+
+We can subscribe to asynchronous changes in the remote state and update the atom value to match.  This can be done using standard React `useEffect()` hook or other popular libraries.
+
+```jsx
+function CurrentUserIDSubscription() {
+  const setCurrentUserID = useSetRecoilState(currentUserIDState);
+
+  useEffect(() => {
+    RemoteStateAPI.subscribeToCurrentUserID(setCurrentUserID);
+    // Specify how to cleanup after this effect
+    return function cleanup() {
+      RemoteServerAPI.unsubscribeFromFriendStatus(setCurrentUserID);
+    };
+  }, []);
+
+  return null;
+}
+
+function MyApp() {
+  return (
+    <RecoilRoot>
+      <CurrentUserIDSubscription />
+      <CurrentUserInfo />
+    </RecoilRoot>
+  );
+}
+```
+
+If you want to handle synchronization of multiple atoms in a single place, you can also use the [State Persistence](persistence) pattern.
+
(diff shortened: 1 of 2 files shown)


docs/docs/guides/asynchronous-state-sync.md
```diff
@@ -2,3 +2,185 @@
...
+If you want to handle synchronization of multiple atoms in a single place, you can also use the [State Persistence](persistence) pattern.
+
+## Bi-Directional Synching
+
+You can also sync the state so local changes are updated on the server:
+
+```jsx
+function CurrentUserIDSubscription() {
+  const [currentUserID, setCurrentUserID] = useRecoilState(currentUserIDState);
+  let knownServerCurrentUserID;
```
acutmore, line 62: `knownServerCurrentUserID` should be stored in `React.useState` so that changes to it will be able to trigger the useEffect

docs/docs/guides/asynchronous-state-sync.md
```diff
@@ -2,3 +2,185 @@
...
+const friendStatusState = atomFamily({
+  key: 'Friend Status',
+  default: 'offline',
+});
+
+function useFriendStatusSubscription(id) {
+  const setStatus = useSetRecoilState(friendStatusState(id));
+
+  useEffect(() => {
+    RemoteStateAPI.subscribeToFriendStatus(id, setStatus);
```
acutmore, line 104: Is it worth making it explicitly clear that it should be taken into consideration that if this hook is used in multiple places, each instance will separately create a subscription?

docs/docs/guides/asynchronous-state-sync.md
```diff
@@ -2,3 +2,185 @@
...
+    }
+
+    RemoteStateAPI.subscribeToCurrentUserID(handleUserChange);
+    // Specify how to cleanup after this effect
+    return function cleanup() {
+      RemoteServerAPI.unsubscribeFromFriendStatus(handleUserChange);
+    };
+  }, []);
+
+  // Subscribe atom changes to update server state
```
acutmore, line 78: I wonder if it's worth making this a little bit more robust, e.g. by including timestamp comparisons, or version counters.
Right now I think there is a race condition if the Recoil state is updated at roughly the same time that the remote values changes.

docs/docs/guides/asynchronous-state-sync.md
drarmstr: I updated it with `useRef`; what I get for writing at 3am.. ;)

docs/docs/guides/asynchronous-state-sync.md
drarmstr: Yeah, I can add a note.  So far, syncing remote mutable state via Recoil isn't a common use-case internally, so we can do some more iteration to provide a more robust example/approach for this use-case.  It would be nice to incorporate more of a singleton pattern here.  But, as also mentioned, lifetimes of components using a friend directly may not match lifetimes of selectors using a friend.  The cleaner answer here may also be related to async query cancelation and memory management for atom/selector lifetime.  If we have an API for that we could perhaps use it for this as well.

docs/docs/guides/asynchronous-state-sync.md
drarmstr: It also suffers from redundant setting of the local atom if the user sets it and the server mirrors back the change to all subscribers, though the value should be correct.  Yeah, just wanted a simplified example, but I'm not very keen on it so far.  Could definitely be improved.

drarmstr:
@acutmore - some minor tweaks in #173

PR #175, opened 2020-05-27
add chinese Translation

stfenjobs:
add documents for chinese language


mondaychen:
We strongly recommend sending PR to docs/docs_zh-cn instead of keeping it within docschina.

PR #195, opened 2020-05-28
fix files config in package.json

kentcdodds:
With the new build, it's now generating multiple files in the dist which should be included in the npm package.


mondaychen:

kentcdodds:
I don't think it was. That updated a different file: `src/package.json`

kentcdodds:
Looks like this will be handled in #200

PR #199, opened 2020-05-29
Fix cannot destructure property error of undefined

jonthomp:
Adds a default value `{}` to the 3rd argument of `recoverableViolation` so that it has something to try and destructure `error` from when nothing is passed at call-sites.

Note: the third argument isn't currently used anywhere in this codebase


alexandrzavalii:
maybe small unit test for it?

PR #200, opened 2020-05-29
babel plugins consistency

### LINKED ISSUE #198: Recoil utils export
I realize this is still upcoming and I'm not asking for a timeline or anything. I'm just creating some educational material on this and want to make sure that I'm making the correct assumptions.

All I want to know is that when the utils like `atomFamily` are exported, how will that be done?

```javascript
// option 1 (my personally preferred option)
import {atomFamily} from 'recoil'

// option 2
import {utils} from 'recoil'
const {atomFamily} = utils

// option 3
import {atomFamily} from 'recoil/utils'

// something else?
```

I realize that nobody can make any guarantees until it's actually released, but if someone can give me any idea of what it might be like that would be helpful :)

jaredpalmer:
Fix what is broken and blocking a new release.

- Fix missing plugin in `.babelrc`
- Sort and ensure that babel plugins are same in both .babelrc and rollup
- Add `preset-fbjs/plugins/dev-expression` to strip `__DEV__` code from prod bundle and remove env checks in dev. We use this on TSDX to shave bytes https://github.com/jaredpalmer/tsdx#advanced-babel-plugin-dev-expressions 

The tests are failing legitimately AFAICT. Wondering if this is because it uses an either some kind of special fbjs setup environment, or perhaps the www build of react? I tried against react@experimental and still failing. I'm pretty sure it's not a jest or babel issue at the moment. 

However, these changes allow for a new 0.0.8 release, fixing tests can come later. 

Closes #198 #197 #107 #33


drarmstr:
cc @mondaychen

src/Recoil_index.js
```diff
@@ -76,4 +92,20 @@ module.exports = {
...
+  readOnlySelector,
+
+  // Concurrency Helpers
+  noWait,
+  waitForNone,
+  waitForAny,
+  waitForAll,
+
+  // Components
+  Link_UNSTABLE: Link,
```
mondaychen: We've decided not to export this one yet.

jaredpalmer:
K @mondaychen this is now ready.

PR #205, opened 2020-05-30
Throw error explicitly if atom is not provided.

alexandrzavalii:
When you call the hook without an atom/selector. `useRecoilState()`
It will throw the error `Error: Cannot read property 'key' of undefined`

Maybe we could check if the atom or selector is provided, if not then throw a meaningful error.
And the check could be applied to all other hooks as well.

Please let me know what do you think. :)


src/hooks/Recoil_Hooks.js
```diff
@@ -232,6 +233,7 @@ function useInterface(): RecoilInterface {
     function useRecoilState<T>(
       recoilState: RecoilState<T>,
     ): [T, SetterOrUpdater<T>] {
+      if (!isRecoilValue(recoilState)) throw new Error('Invalid atom provided');
```
drarmstr, line 236: nodes may be either atoms or selectors.  the type is `RecoilState<T>`

drarmstr:
We should handle all hooks consistently.

cc @habond - What we were discussing..

src/hooks/Recoil_Hooks.js
alexandrzavalii: yes, isRecoilValue should be checking for RecoilState.
Maybe the error message should be? `No atom or selector provided`

alexandrzavalii:
@drarmstr yes we should add the similar check to all other hooks.

src/hooks/Recoil_Hooks.js
```diff
@@ -232,6 +233,8 @@ function useInterface(): RecoilInterface {
     function useRecoilState<T>(
       recoilState: RecoilState<T>,
     ): [T, SetterOrUpdater<T>] {
+      if (!isRecoilValue(recoilState))
+        throw new Error('No atom or selector provided');
```
acutmore, line 237: Perhaps these checks could be wrapped in a check for `__DEV__` to keep the bundle size down in production.

alexandrzavalii: @acutmore good idea!

alexandrzavalii:
@drarmstr @acutmore Thanks for taking a look at it! :) 
Let me know if the check looks okay, and I will add it to all the hooks.

src/hooks/Recoil_Hooks.js
```diff
@@ -232,6 +233,11 @@ function useInterface(): RecoilInterface {
     function useRecoilState<T>(
       recoilState: RecoilState<T>,
     ): [T, SetterOrUpdater<T>] {
+      if (__DEV__) {
+        if (!isRecoilValue(recoilState))
```
drarmstr, line 237: Looks like we don't have prettier running automatically on the GitHub side.  All if statements need blocks, single-statements aren't allowed.

src/hooks/Recoil_Hooks.js
```diff
@@ -232,6 +233,11 @@ function useInterface(): RecoilInterface {
...
+          throw new Error('No atom or selector provided');
```
drarmstr, line 238: Hmm... maybe not clear that what was passed was invalid vs needing something else to be passed in - maybe "Invalid parameter, atom or selector expected"

src/hooks/Recoil_Hooks.js
```diff
@@ -208,12 +218,18 @@ function useInterface(): RecoilInterface {
...
+      }
       return () => setRecoilValue(storeRef.current, recoilState, DEFAULT_VALUE);
     }
 
     function useRecoilValueLoadable<T>(
       recoilValue: RecoilValue<T>,
     ): Loadable<T> {
+      if (__DEV__) {
+        validateRecoilState(recoilState);
+      }
```
alexandrzavalii, line 232: `useRecoilValueLoadable` is used by `useRecoilValue` which is used by `useRecoilState`

drarmstr:
Local edits on Phabricator:
* Name `validateRecoilValue()` for type consistency
* Fix `useRecoilValueLoadable` to pass `recoilValue` instead of `recoilState`
* Remove references to `console`
* Update test to use custom component that catches error to check if contents have the word parameter to see if it has a meaningful message since an error will be thrown either way.
* Misc lint

davidmccabe:
This is a good change. I would prefer if the message was: `Invalid argument to ${hookName}: expected an atom or selector but got ${serializedParameter}.` as this explains what's going on a bit more thoroughly and would make debugging the problem a little faster. Thanks so much for contributing this!

drarmstr:
Updated error message on Phabricator, no need to update PR.

PR #220, opened 2020-05-31
Adds window object check in Recoil Root

### LINKED ISSUE #215: Recoil breaks w/ Next.js (window undefined on server)
Decided to play around with setting up recoil w/ next.js. The issue is it looks like in development mode, recoil looks for the window object which is undefined when next renders the app on the server.

Digging a little deeper, it stems from [these](https://github.com/facebookexperimental/Recoil/blob/52c3adf447e4f32c16e7e086a553f20d9ea743d7/src/components/Recoil_RecoilRoot.react.js#L123) window calls in dev mode.

For the time being, I created a workaround that I came up with [here](https://github.com/Chrischuck/next-recoil-setup/tree/master/src) essentially loads RecoilRoot(and its children) on the client only.

Would be happy to add a check to see if window is defined before calling any methods on it as well as diving into any other issues that might come up when combing recoil with next :)

Chrischuck:
Should fix #215

In short, it checks if window is defined before using it in dev mode.

I don't think these minimal changes should brea but I haven't tested them.

This is the first non-UI library I've worked on and I don't know how to test my changes locally (other than seeing if build + tests pass or building and using it in an app).

If someone could let me know the preferred way to test that would be awesome!

### DIFF
--- a/src/components/Recoil_RecoilRoot.react.js
+++ b/src/components/Recoil_RecoilRoot.react.js
@@ -213,8 +213,10 @@ function RecoilRoot({initializeState, children}: Props): ReactElement {
       return;
     }
     // Save changes to nextTree and schedule a React update:
-    if (__DEV__ && typeof window !== 'undefined') {
-      window.$recoilDebugStates.push(replaced); // TODO this shouldn't happen here because it's not batched
+    if (__DEV__) {
+      if (typeof window !== 'undefined') {
+        window.$recoilDebugStates.push(replaced); // TODO this shouldn't happen here because it's not batched
+      }
     }
     storeState.nextTree = replaced;
     nullthrows(notifyBatcherOfChange.current)();
--- a/src/components/Recoil_RecoilRoot.react.js
+++ b/src/components/Recoil_RecoilRoot.react.js
@@ -120,7 +120,7 @@ function Batcher(props: {setNotifyBatcherOfChange: (() => void) => void}) {
 }
 
 if (__DEV__) {
-  if (!window.$recoilDebugStates) {
+  if (typeof window !== 'undefined' && !window.$recoilDebugStates) {
     window.$recoilDebugStates = [];
   }
 }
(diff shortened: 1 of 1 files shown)


src/components/Recoil_RecoilRoot.react.js
```diff
@@ -120,7 +120,7 @@ function Batcher(props: {setNotifyBatcherOfChange: (() => void) => void}) {
 }
 
 if (__DEV__) {
-  if (!window.$recoilDebugStates) {
+  if (typeof window !== 'undefined' && !window.$recoilDebugStates) {
```
Chrischuck, line 123: Should we make this a single if?

```
if (__DEV__ && typeof...)
```

drarmstr: @mondaychen - for the sync scripts does the `__DEV__` check need to be by itself?

src/components/Recoil_RecoilRoot.react.js
```diff
@@ -213,7 +213,7 @@ function RecoilRoot({initializeState, children}: Props): ReactElement {
       return;
     }
     // Save changes to nextTree and schedule a React update:
-    if (__DEV__) {
+    if (__DEV__ && typeof window !== 'undefined') {
```
drarmstr, line 216: Can we use a standard `window != null` check?

src/components/Recoil_RecoilRoot.react.js
mondaychen: No. `__DEV__` can be used like a simple boolean variable.

drarmstr: FYI - We have an internal lint warning to keep `__DEV__` checks on their own line for clarity since they are crushed in production.

drarmstr:
Making edits locally in Phabricator

src/components/Recoil_RecoilRoot.react.js
mondaychen: Cool. What I said only applies to our github build process.

src/components/Recoil_RecoilRoot.react.js
Chrischuck: I believe `typeof` is the only way to check if the window object is valid on the server since `window` is never declared, checking if it's `null` would still error out

dustinsoftware:
Chris is correct, you can try it out in the Node REPL to confirm. ‘typeof
window’ will prevent throwing errors in strict mode since Node environments
won’t have it defined at all.

Chrischuck:
Reason for closing?

PR #224, opened 2020-06-01
Fix asynchronous-data-queries.md to use atomFamily instead of selectorFamily

alexandrzavalii:
Correct me if I am wrong, but in this example `get` from `selectorFamily` is not used, so we could just use `atomFamily` instead.


alexandrzavalii:
just realized, atomFamily does not have `get` request

PR #232, opened 2020-06-01
Update Guides (Queries with Parameters)  use atomFamily instead of selectorFamily

alexandrzavalii:
I don't see the reason of using `selectorFamily` in this example.


drarmstr:
Atoms represent and maintain state while selectors represent functions.  Using atoms here would introduce unnecessary state which is not used.  

Atoms with asynchronous defaults are actually implemented by wrapping with a selector.  Using an atomFamily like this would still really be using selectors along with the unused atom.

alexandrzavalii:
@drarmstr thanks! now makes sense. :)

PR #233, opened 2020-06-02
fix: Cannot be used in the DEV environment of the SSR project

fyber-LJX:
I got an error when I use Recoil in the development environment of next.js project，this error is "ReferenceError: window is not defined", so I want to fix this problem.


drarmstr:
Duplicate of #220

PR #240, opened 2020-06-02
Enable RSS & Atom feeds for blog

renchap:
According to the documentation (https://v2.docusaurus.io/docs/blog#feed) only a small config update is needed.


davidmccabe:
Thanks!

PR #243, opened 2020-06-03
Fixed typo

rajanmali:
useRecoilState() -> useRecoilSetState()


drarmstr:
@rajanmali  - This isn't a typo.  If `useSetRecoilState()` were used then we avoid a subscription.  This is calling out that if we used `useRecoilState()` to obtain the setter via the second entry in the array return value then it would also subscribe the component to that state.

PR #249, opened 2020-06-05
Add react and react-dom as dev dependencies

mondaychen:
This is necessary for tests to run. Currently when running `yarn test`, many files cannot even load because "cannot resolve modules 'react'". After adding these as devDependencies, you still see some errors, but as least from actual tests.


dustinsoftware:
Should install-peers-cli be dropped as a dependency then?

PR #254, opened 2020-06-05
Using rollup-plugin-copy instead of postbuild script

wilbert-abreu:


dustinsoftware:
Yeah when I implemented this I opted to use a small node script because of how many additional deps that get pulled in. Are you seeing issues with the current script setup that this PR addresses?

wilbert-abreu:
hey @dustinsoftware, thanks for your question and for implementing that. No current issue, but I was going to suggest adding a watch script via the `-w` flag for local dev and realized having to run the script each time after each build would complicate things. 

Is there a downside to adding some additional dev deps?

mondaychen:
This has merge conflicts with #292. Whichever merged later should fix some content accordingly.

wilbert-abreu:
Thanks @mondaychen, I solved the merge conflicts

mondaychen:
@wilbert-abreu Hi. Please remove postbuild.js and add `typescript/index.d.ts` to copy targets. Thanks!

mondaychen:
Closing since `postbuild.js` is now serving different purpose.

PR #267, opened 2020-06-07
[docs] Improve API Reference documentation

drarmstr:


docs/docs/api-reference/core/selector.md
```diff
@@ -3,19 +3,47 @@ title: selector(options)
...
+
+  get: ({
+    get: GetRecoilValue
+  }) => T | Promise<T> | RecoilValue<T>,
+
+  set?: (
+    {
+      get: GetRecoilValue,
+      set: SetRecoilValue,
+      reest: ResetRecoilValue,
```
wilbert-abreu, line 22: hey @drarmstr, slight typo here fyi `reest` instead of `reset`

drarmstr: Thanks, that was fixed in #275

PR #274, opened 2020-06-07
Fix selector() options "set: SetRecoilValue" type to "SetRecoilState"

dharkness:
The setter function is defined as SetRecoilState, but the selector's
options.set type is SetRecoilValue which I do not see defined anywhere.

Note: I still feel SetRecoilValue is better as it matches what it does
and mirrors GetRecoilValue. Additionally, the useSetRecoilState() hook
should be renamed to useSetRecoilValue() to match useRecoilValue().

Given the typo this commit fixes, I am not alone. :)

### DIFF
--- a/docs/docs/api-reference/core/selector.md
+++ b/docs/docs/api-reference/core/selector.md
@@ -18,7 +18,7 @@ function selector<T>({
   set?: (
     {
       get: GetRecoilValue,
-      set: SetRecoilValue,
+      set: SetRecoilState,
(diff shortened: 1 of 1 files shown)


drarmstr:
The reason they are typed differently is that they have different restrictions on the nodes they accept:

* `RecoilState` - Writeable value
* `RecoilValueReadOnly` - Read-only value
* `RecoilValue` - Read-only or Writeable value

PR #278, opened 2020-06-07
[docs] Fix a couple issues in persistence.md

### LINKED ISSUE #269: Not possible to restore atoms created with atomFamily
given the atomFamily:

```
const testItem = atomFamily<number>({
  key: 'testItem',
  default: null
});
```

creating an item with `set(testItem("myId"), 123)` triggers a call to 
```
 useTransactionObservation_UNSTABLE({atomValues, atomInfo, modifiedAtoms} => {
      const list = [];
      for (const modifiedAtom of modifiedAtoms) {
        list.push({name: modifiedAtom, value: JSON.stringify(atomValues.get(modifiedAtom))});
      }
    saveStore(list)
 });
```
with the key as `testItem__"myId"`. When trying to restore that with 
```
set({key: item.name}, JSON.parse(item.value));
```
it fails with error `Missing definition for RecoilValue: "testItem__"myId""`

Also `atomValues.get(modifiedAtom)` always return nothing as atomValues is always empty.

Also there is an error in the documentation that says we should use `modifiedAtom.key`. `modifiedAtom` is a string though and therefore does not have `key` property.

drarmstr:
Resolves #269

### DIFF
--- a/docs/docs/guides/persistence.md
+++ b/docs/docs/guides/persistence.md
@@ -22,7 +22,7 @@ function PersistenceObserver() {
   useTransactionObservation_UNSTABLE(({atomValues, atomInfo, modifiedAtoms}) => {
     for (const modifiedAtom of modifiedAtoms) {
       Storage.setItem(
-        modifiedAtom.key,
+        modifiedAtom,
         JSON.stringify({value: atomValues.get(modifiedAtom)}),
       );
     }
@@ -54,7 +54,7 @@ Here is a basic example:
 ```jsx
 const initializeState = ({set}) => {
   for(const [key, value] of Storage.entries()) {
-    set({key}, JSON.parse(value)).value;
+    set(getAtomWithKey(key), JSON.parse(value)).value;
   }
 }
 


docs/docs/guides/persistence.md
```diff
@@ -54,7 +54,7 @@ Here is a basic example:
 ```jsx
 const initializeState = ({set}) => {
   for(const [key, value] of Storage.entries()) {
-    set({key}, JSON.parse(value)).value;
+    set(getAtomWithKey(key), JSON.parse(value)).value;
```
aulneau, line 57: @drarmstr where does `getAtomWithKey` come from?

drarmstr: Some user-defined function.  We hesitated to even document examples for this API as it is changing.  `set()` is difficult to use here for dynamically generated nodes such as `atomFamily()` as you may not have them defined at initialization time (See #269).  Internally we use `setUnvalidatedAtomValues()`, which takes a key and does not require the atom to be defined yet.  But, that requires additional metadata in the atom to work.  You can play with it now if you like, or stay tuned for the updated API which cleans all this up.

PR #284, opened 2020-06-08
fix: add window and dev check in stableStringify

aulneau:
While playing around with Recoil this weekend, I tried to use `selectorFamily` but it seems to be using `stableStringify` which has an unchecked window call, and with something like next.js it was causing errors on the server. This fixes it!

### DIFF
--- a/src/util/Recoil_stableStringify.js
+++ b/src/util/Recoil_stableStringify.js
@@ -126,23 +126,29 @@ function stringify(x: mixed, opt: Options, key?: string): string {
 // * We could support BigInt, but Flow doesn't seem to like it.
 // See Recoil_stableStringify-test.js for examples
 function stableStringify(x: mixed, opt?: Options): string {
-  const startTime = window.performance ? window.performance.now() : 0;
-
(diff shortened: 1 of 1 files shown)


aulneau:
@drarmstr done!

src/util/Recoil_stableStringify.js
```diff
@@ -126,22 +126,28 @@ function stringify(x: mixed, opt: Options, key?: string): string {
 // * We could support BigInt, but Flow doesn't seem to like it.
 // See Recoil_stableStringify-test.js for examples
 function stableStringify(x: mixed, opt?: Options): string {
+  const shouldEvalPerformance = __DEV__ && typeof window !== 'undefined';
```
drarmstr, line 129: Sorry, for the code builds to work fully they require `if (__DEV__)` checks to be their own statements.

drarmstr:
> @drarmstr done!

@aulneau Thanks so much for helping with this!

src/util/Recoil_stableStringify.js
aulneau: oh ok, np. How does my latest push look?

PR #292, opened 2020-06-09
TypeScript types and associated tests

### LINKED ISSUE #291: Move TypeScript definitions into core package
Hello all!

We've heard your feedback on TS types, and we've decided to migrate TypeScript types to the core package and keep it in sync with API changes. Ultimately that means you won't have to install the `@types/recoil` dependency anymore to access TS types, and you won't have to worry about mismatched versions or missing types.

Right now, the types live in DefinitelyTyped, so I'll be removing the types on that repo and opening a PR in this repo with the types to get us all set up.

csantos42:
To run tests:

```
npm run test:typescript
```

fixes #291

### DIFF
--- a/scripts/postbuild.js
+++ b/scripts/postbuild.js
@@ -1,3 +1,10 @@
 const fs = require('fs');
 
-fs.copyFileSync('./src/npm/index.js', './dist/index.js');
+function logErrors(err) {
+  if (err) {
+    throw err;
+  }
+}
+
+fs.copyFile('./src/npm/index.js', './dist/index.js', logErrors);
+fs.copyFile('./typescript/index.d.ts', './dist/index.d.ts', logErrors);
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -139,7 +139,6 @@ export function useRecoilCallback<Args extends ReadonlyArray<unknown>, Return>(
 // loadable.d.ts
 type ResolvedLoadablePromiseInfo<T> = Readonly<{
   value: T;
-  upstreamState__INTERNAL_DO_NOT_USE?: TreeState;
 }>;
 
 export type LoadablePromise<T> = Promise<ResolvedLoadablePromiseInfo<T>>;
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -1,0 +1,264 @@
+// Minimum TypeScript Version: 3.7
+
+// state.d.ts
+export type NodeKey = string;
+export type AtomValues = Map<NodeKey, Loadable<any>>;
+export type ComponentCallback = (state: TreeState) => void;
+export type TreeState = Readonly<{
+    isSnapshot: boolean;
+    transactionMetadata: object;
+    dirtyAtoms: Set<NodeKey>;
+    atomValues: AtomValues;
+    nonvalidatedAtoms: Map<NodeKey, unknown>;
+    nodeDeps: Map<NodeKey, Set<NodeKey>>;
+    nodeToNodeSubscriptions: Map<NodeKey, Set<NodeKey>>;
+    nodeToComponentSubscriptions: Map<NodeKey, Map<number, [string, ComponentCallback]>>;
+}>;
+
+// node.d.ts
+export class DefaultValue {}
+
+// recoilRoot.d.ts
+import * as React from 'react';
+
+export interface RecoilRootProps {
+    initializeState?: (options: {
+        set: <T>(recoilVal: RecoilState<T>, newVal: T) => void;
+        setUnvalidatedAtomValues: (atomMap: Map<string, unknown>) => void;
+    }) => void;
+}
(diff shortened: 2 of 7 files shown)


mondaychen:
This is awesome!
Is it a convention in typescript world that we use 4 space indent?

csantos42:
> This is awesome!
> Is it a convention in typescript world that we use 4 space indent?

ah good catch, they require it in `DefinitelyTyped`, but we should be able to override that now that we own the files-- I'll make that change

typescript/index.d.ts
```diff
@@ -0,0 +1,264 @@
...
+
+// node.d.ts
+export class DefaultValue {}
+
+// recoilRoot.d.ts
+import * as React from 'react';
+
+export interface RecoilRootProps {
+  initializeState?: (options: {
+      set: <T>(recoilVal: RecoilState<T>, newVal: T) => void;
```
ryprice, line 26: nit a few lingering 4 spaces near the top 😃

mondaychen:
Hey @csantos42 I think you forgot to commit the changes in yarn.lock

typescript/index.d.ts
```diff
@@ -0,0 +1,264 @@
...
+ */
+export function useRecoilCallback<Args extends ReadonlyArray<unknown>, Return>(
+  fn: (interface: CallbackInterface, ...args: Args) => Return,
+  deps?: ReadonlyArray<unknown>,
+): (...args: Args) => Return;
+
+// loadable.d.ts
+export type ResolvedLoadablePromiseInfo<T> = Readonly<{
+  value: T;
+  upstreamState__INTERNAL_DO_NOT_USE?: TreeState;
```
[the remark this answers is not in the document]
csantos42: good point! I'll go through file and remove anything that I don't think will be useful to export for users

typescript/index.d.ts
csantos42: ah good catch! I'll remove

csantos42:
ohh woops @mondaychen I'll make that change thanks!

typescript/index.d.ts
```diff
@@ -0,0 +1,264 @@
...
+
+// bigint not supported yet
+export type Primitive = undefined | null | boolean | number | symbol | string;
+
+export type SerializableParam = Primitive | SerializableParam[] | { [key: string]: SerializableParam };
+
+export interface AtomFamilyOptions<T, P extends SerializableParam> {
+  key: NodeKey;
+  dangerouslyAllowMutability?: boolean;
+  default: RecoilValue<T> | Promise<T> | T | ((param: P) => T | RecoilValue<T> | Promise<T>);
```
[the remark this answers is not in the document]
csantos42: good point! But I should've mentioned we don't want to introduce any new terminology in this file so that the source of truth lives in flow and this just mirrors that implementation as much as possible (I'll add a comment at the top of the file to ensure that's the case)

typescript/index.d.ts
```diff
@@ -0,0 +1,264 @@
...
+}
+
+export function atomFamily<T, P extends SerializableParam>(
+  options: AtomFamilyOptions<T, P>,
+): (param: P) => RecoilState<T>;
+
+export interface ReadOnlySelectorFamilyOptions<T, P extends SerializableParam> {
+  key: string;
+  get: (param: P) => (opts: { get: GetRecoilValue }) => Promise<T> | RecoilValue<T> | T;
```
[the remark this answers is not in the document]
csantos42: good point! But I should've mentioned we don't want to introduce any new terminology in this file so that the source of truth lives in flow and this just mirrors that implementation as much as possible (I'll add a comment at the top of the file to ensure that's the case)

csantos42:
thanks @mondaychen now we should be good to go :)

scripts/postbuild.js
```diff
@@ -1,3 +1,10 @@
...
 
-fs.copyFileSync('./src/npm/index.js', './dist/index.js');
+function logErrors(err) {
+  if (err) {
+    throw err;
+  }
+}
+
+fs.copyFile('./src/npm/index.js', './dist/index.js', logErrors);
+fs.copyFile('./typescript/index.d.ts', './dist/index.d.ts', logErrors);
```
ayush987goyal, line 10: I think this should be 
```ts
fs.copyFile('./typescript/index.d.ts', './dist/typescript/index.d.ts', logErrors);
```
Since in `package.json` we are mentioning `files` as `dist` and `types` as `typescript` directory. Hence we would need to have same directory structure inside `dist`.

csantos42: hey! Good catch-- I meant to remove the `types` property as we are just putting `index.d.ts` at the top level. Thanks!

MartynasZilinskas:
@csantos42 missing `types` property in package.json.
It should be `"types": "dist/index.d.ts"`

PR #294, opened 2020-06-09
Fix broken async related tests

mondaychen:
Current master:
Tests:       42 failed, 113 passed, 155 total

This branch:
Test Suites: 4 failed, 16 passed, 20 total
Tests:       7 failed, 148 passed, 155 total

Two things this PR does:
- Adding 'promise-polyfill' as dev dependency use `setTimeout` based `Promise` instead of native to mock FB internal behavior. This will allow `jest.runAllTimers()` flush Promises as well. 
- Some manual fixes on Recoil_selector-test.js (additional `await` on async selectors)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "jest-cli": "^26.0.1",
     "lint-staged": ">=10",
     "prettier": "^2.0.5",
+    "promise-polyfill": "^8.1.3",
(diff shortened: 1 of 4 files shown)


aaronabramov:
oh.. so the custom FB promises were the issue? that makes sense. i'll close my PR from last week then :)

PR #308, opened 2020-06-11
Improve flow support

Komalov:
### What is this
It has 150+ errors related to flow right now. Some of them can be fixed without any noticeable changes so it can be suppressed to just 28 errors easily

### DIFF
--- a/src/adt/Recoil_Loadable.js
+++ b/src/adt/Recoil_Loadable.js
@@ -32,7 +32,7 @@ type ResolvedLoadablePromiseInfo<+T> = $ReadOnly<{
 
 export type LoadablePromise<T> = Promise<ResolvedLoadablePromiseInfo<T>>;
 
-type Accessors<T> = $ReadOnly<{
+type Accessors<T> = $ReadOnly<{|
   // Attempt to get the value.
   // If there's an error, throw an error.  If it's still loading, throw a Promise
   // This is useful for composing with React Suspense or in a Recoil Selector.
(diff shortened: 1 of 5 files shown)


aaronabramov:
this is great! thanks for working on this!

how hard would it be to fix those 28 errors? if it's OSS/GH specific, can we add a bunch of $FlowFixMe's and add `yarn flow` to the list of steps on CI?

aaronabramov:
actually nevermind. `yarn flow` command is already there and broken, so this is definitely improving the situation already. let me import it and see if it works with internal flow build as well

Komalov:
It requires more deep understanding of how things work what I currently do not have. But it looks like some stuff more related to flow itself e.g. https://github.com/facebookexperimental/Recoil/blob/c336325c76a68ace6e8da53b28866e442613d855/src/util/Recoil_mergeMaps.js#L22 . Seems flow doesn't notice assignment into the loop condition

src/adt/Recoil_Loadable.js
```diff
@@ -32,7 +32,7 @@ type ResolvedLoadablePromiseInfo<+T> = $ReadOnly<{
 
 export type LoadablePromise<T> = Promise<ResolvedLoadablePromiseInfo<T>>;
 
-type Accessors<T> = $ReadOnly<{
+type Accessors<T> = $ReadOnly<{|
```
aaronabramov, line 35: seems like objects are now exact by default, is it not the case in this configuration?

Komalov: I didn't know that, but sure enough I needed to make exact types here manually. Need to check if [utility type](https://flow.org/en/docs/types/utilities/#toc-readonly) behaviours differently

PR #318, opened 2020-06-12
Reduce public flow errors

Komalov:
Some fixes and workarounds that helps to pass flow CI

### DIFF
--- a/src/adt/Recoil_ArrayKeyedMap.js
+++ b/src/adt/Recoil_ArrayKeyedMap.js
@@ -27,6 +27,7 @@ class ArrayKeyedMap<V> {
   constructor(
     existing?: ArrayKeyedMap<V> | Iterable<[mixed, V]>,
   ): ArrayKeyedMap<V> {
+    // $FlowOSSFixMe
     this._base = new Map(); // @oss-only
 
     if (existing instanceof ArrayKeyedMap) {
@@ -43,6 +44,7 @@ class ArrayKeyedMap<V> {
 
   get(key: mixed): V | void {
     const ks = Array.isArray(key) ? key : [key];
+    // $FlowOSSFixMe
     let map = this._base;
     ks.forEach(k => {
       map = map.get(k) ?? emptyMap;
@@ -52,6 +54,7 @@ class ArrayKeyedMap<V> {
 
   set(key: mixed, value: V): any {
     const ks = Array.isArray(key) ? key : [key];
+    // $FlowOSSFixMe
     let map = this._base;
     let next = map;
     ks.forEach(k => {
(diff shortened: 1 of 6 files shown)


src/adt/Recoil_ArrayKeyedMap.js
```diff
@@ -27,6 +27,7 @@ class ArrayKeyedMap<V> {
   constructor(
     existing?: ArrayKeyedMap<V> | Iterable<[mixed, V]>,
   ): ArrayKeyedMap<V> {
+    // $FlowFixMe
```
aaronabramov, line 30: this will result in `unused suppression` warning internally.
can you add `$FlowOSSFixMe` regex to `.flowconfig` and use that instead?

src/recoil_values/Recoil_atomFamily.js
```diff
@@ -177,18 +177,18 @@ function atomFamily<T, P: Parameter>(
   let legacyAtom;
   // prettier-ignore
   // @fb-only: if (
-    // @fb-only: options.scopeRules_APPEND_ONLY_READ_THE_DOCS
+  // @fb-only: options.scopeRules_APPEND_ONLY_READ_THE_DOCS
```
aaronabramov, line 180: unfortunately these are generated by the build system that is patching the code while syncing it to github and should probably not be touched. do you mind reverting the formatting?

src/recoil_values/Recoil_atomFamily.js
```diff
@@ -235,10 +235,7 @@ function atomFamily<T, P: Parameter>(
     const newAtom = atom<T>({
       key: `${options.key}__${stableStringify(params) ?? 'void'}`,
       default: atomFamilyDefault(params),
-      scopeRules_APPEND_ONLY_READ_THE_DOCS: mapScopeRules(
-        options.scopeRules_APPEND_ONLY_READ_THE_DOCS,
-        params,
-      ),
+      // @fb-only: scopeRules_APPEND_ONLY_READ_THE_DOCS: mapScopeRules(options.scopeRules_APPEND_ONLY_READ_THE_DOCS, params),
```
aaronabramov, line 238: sorry.. i don't think this would work either 😔 
`// @fb-only` comments are actually generating during the sync build and aren't stored internally
@mondaychen @drarmstr do you know what the process is usually for these when coming from external contributions?

aaronabramov: actually it seems like the system is smart enough to do the same sync in revers order :) 
can you fix the formatting so it doesn't change the lines?
Screen Shot 2020-06-12 at 5 45 10 PM

src/adt/Recoil_ArrayKeyedMap.js
Komalov: Sure

src/recoil_values/Recoil_atomFamily.js
aaronabramov: oh... it even fixed the formatting internally after rebasing. i think this should be good then. @Komalov update the diff and i'll re-import it again

PR #329, opened 2020-06-14
initial set up for eslint

strangeforloop:


.flowconfig
```diff
@@ -43,4 +43,4 @@ unsafe-getters-setters
 all=warn
 
 [version]
-^0.124.0
+^0.124.0
```
aaronabramov, line 46: you'll probably need to change this to `0.126.0`

also you should configure your editor to add a newline at the end of the file
https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files#:~:text=Additionally%2C%20in%20the%20Settings%20page,there%20isn't%20already%20one.

package.json
```diff
@@ -73,5 +77,8 @@
   },
   "lint-staged": {
     "*.{js,md,json}": "prettier --write"
+  },
+  "dependencies": {
+    "eslint-plugin-react": "^7.20.0"
```
aaronabramov, line 82: can you move it from "dependencies" to "devDepenedcies"?

.flowconfig
strangeforloop: I'll address both of these now.

PR #343, opened 2020-06-17
Remove duplicate flow-bin dependency

jacques-blom:
There is a duplicate entry for `flow-bin` in the `package.json`. This PR removes it.


drarmstr:
Duplicate of #339 

Thanks for fixing this! but, should already be landing with #339

Issue #228, opened 2020-06-01
Performance bug?


Ruthenz:
Hey,

I'm not sure if this is a performance bug, or maybe too many atoms just overloads the system but I created a simple example of a Canvas component with width and height that each point is a Pixel component that gets it's state from an atom (with a specific id).

It works great with 20x20 canvas, a bit slow with 50x50 and just crashes completely with 100x100.

I created a sandbox example : https://codesandbox.io/s/crimson-thunder-e45pj?file=/src/Canvas.tsx

You can change the width and height and see what happens.

Would love to understand why this happens and if I'm doing something wrong :)

acutmore:
Hi @Ruthenz.

I can't see anything specifically that you're doing wrong with Recoil. Only the affected Atoms are being triggered. I tried your setup with React's `useState` instead of `useRecoilState` and the performance was still not great.

I think it's the DOM updates which are the slowest part here. All the `<div>` are positioned with flex. I would try absolute positioning each pixel so the browser has less work to do.

Ruthenz:
The problem was that the app completely crashes as opposed to working slow.
I also tried using React hooks and the performance was terrible (even with 50x50 compared to Recoil's much better performance), but it didn't crash.

So I wondered if it has something to do with too many atoms.

acutmore:
Hi @Ruthenz 

Yes I think you're right. The crash is being caused by Recoil. I downloaded the sandbox so it would be easier to inspect in devtools and am seeing an out-of-memory exception happening from Recoil's `subscribeComponentToNode`.

jwdinker:
@Ruthenz I can't mess with it because my browser is freezing, but from a glance have you tried memozing your onMouseEnter callback instead of an anon fn?  You are potentially creating and destroying tens of thousands of functions with every pixel change.

JimLiu:
> Hi @Ruthenz
> 
> Yes I think you're right. The crash is being caused by Recoil. I downloaded the sandbox so it would be easier to inspect in devtools and am seeing an out-of-memory exception happening from Recoil's `subscribeComponentToNode`.

Yes, you are right👍, it's caused by `subscribeComponentToNode`, the root reason is `mapByUpdatingInMap`
```js
function mapByUpdatingInMap(map, k, updater) {
  const next = new Map(map);
  next.set(k, updater(next.get(k)));
  return next;
}
```

[...]

PR #348, opened 2020-06-18
Performance improvement

JimLiu:
This PR relates to #228 

I [forked the code](https://codesandbox.io/s/d302y) from the original post, and changed the pixel count to 100 * 80 * 2, then use the Chrome Dev tool to profile the [page](https://d302y.csb.app/) performance, and you can see it spends long time on `mapByUpdatingInMap` and GC.

image

if we take a deep look with the function, the `new Map` takes time.

image

This PR is trying to reduce the `new Map`, and I tested it with this PR using the [same test code]  (https://codesandbox.io/s/xituo), and tested the [page](https://xituo.csb.app/) performance again, the performance got much better.
image

### DIFF
--- a/src/util/Recoil_CopyOnWrite.js
+++ b/src/util/Recoil_CopyOnWrite.js
@@ -25,24 +25,18 @@ function setByDeletingFromSet<V>(set: $ReadOnlySet<V>, v: V): Set<V> {
   return next;
 }
 
-function mapBySettingInMap<K, V>(
-  map: $ReadOnlyMap<K, V>,
-  k: K,
-  v: V,
-): Map<K, V> {
-  const next = new Map(map);
-  next.set(k, v);
-  return next;
+function mapBySettingInMap<K, V>(map: Map<K, V>, k: K, v: V): Map<K, V> {
+  map.set(k, v);
+  return map;
 }
 
 function mapByUpdatingInMap<K, V>(
-  map: $ReadOnlyMap<K, V>,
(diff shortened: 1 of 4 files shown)


mondaychen:
@davidmccabe

src/util/Recoil_CopyOnWrite.js
```diff
@@ -25,24 +25,18 @@ function setByDeletingFromSet<V>(set: $ReadOnlySet<V>, v: V): Set<V> {
...
 
-function mapBySettingInMap<K, V>(
-  map: $ReadOnlyMap<K, V>,
-  k: K,
-  v: V,
-): Map<K, V> {
-  const next = new Map(map);
-  next.set(k, v);
-  return next;
+function mapBySettingInMap<K, V>(map: Map<K, V>, k: K, v: V): Map<K, V> {
```
mondaychen, line 28: This is no longer "copy on write" after the change... if there's performance concern, maybe we need to re-think about how we manipulate data.

davidmccabe:
Thanks for profiling and contributing this. Unneeded copies are a bottleneck with Recoil currently, and we are working to remove them. 

Due to concurrent mode, we will always need to maintain multiple states at times. We plan to handle this using persistent data structures, and are exploring different options for that.

Other copies are just unneeded defensive coding that we can get rid of. This diff indiscriminately stops doing all copies, whereas we need to carefully prove that each one is unnecessary and make sure it stays that way with future changes.

PR #368, opened 2020-06-19
Fixes lint errors

Komalov:
The only problem of eslint is missing fb-www package. It might be resolved once the corresponding package will be published by adding it to devDeps. Pls let me know if something goes against internal fb build strategy

### DIFF
--- a/src/hooks/__tests__/Recoil_useRecoilCallback-test.js
+++ b/src/hooks/__tests__/Recoil_useRecoilCallback-test.js
@@ -149,13 +149,12 @@ describe('useRecoilCallback', () => {
     // Something with React and Jest requires this extra kick...
     const [Kick, kick] = componentThatReadsAndWritesAtom(kickAtom);
 
-    /* eslint-disable react/jsx-key */
     const container = renderElements([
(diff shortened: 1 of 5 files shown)


src/hooks/__tests__/Recoil_useRecoilCallback-test.js
```diff
@@ -149,11 +149,13 @@ describe('useRecoilCallback', () => {
     // Something with React and Jest requires this extra kick...
     const [Kick, kick] = componentThatReadsAndWritesAtom(kickAtom);
 
+    /* eslint-disable react/jsx-key */
     const container = renderElements([
       <Component />,
       <ReadsAtom atom={anAtom} />,
       <Kick />,
     ]);
+    /* eslint-enable react/jsx-key */
```
aaronabramov, line 158: @Komalov i don't think we use that rule internally and it warns about unused suppression, can we disable it in here as well?

Komalov: Sure, I disabled this rule for public builds

Issue #12, opened 2020-05-14
Warning: To locate the bad setState() At selector


syunto07ka:
A warning is occured where selector() used even though I write a code like one which is in the [official document here](https://recoiljs.org/docs/introduction/getting-started/#selector).

<img width="516" alt="スクリーンショット 2020-05-15 4 55 26" src="https://user-images.githubusercontent.com/37825725/81980341-61841500-9669-11ea-9427-390d04645eca.png">

this is my code

```
import React, { useEffect } from 'react';
import { atom, useRecoilState, useRecoilValue, selector } from 'recoil';

function RecoilApp() {
  useEffect(() => {
      console.log('test');
  });
  return(
    <div>
      Recoil App
      <TextInput />
      <CharCount />
    </div>
  );
}

const textState = atom({key: 'textState', default: ''});

function TextInput() {
  const [text, setText] = useRecoilState(textState);

  const onChange = event => {
    setText(event.target.value);
  }

  return(
    <div>
      <input type="text" value={text} onChange={onChange} />
      <br />
      Echo: {text}
    </div>
  );
};

function CharCount() {
  const count = useRecoilValue(countState);
  return <>Charset Count: {count}</>;
}

const countState = selector({
  key: 'countState',
  get: ({get}) => {
    const text = get(textState);
    return text.length;
  },
});

export default RecoilApp;

```

ecreeth:
I have copied and pasted your code and it works. [See example in codesandbox](https://codesandbox.io/s/flamboyant-burnell-oiboe?file=/src/App.js)

syunto07ka:
Thanks, ecreeth.
As you say that, my code works, but warning is occured in my browser.

Humm, I wonder that the codesandbox occured no warning at console like the image.

dai-shi:
You need to try with react v16.13.1.
https://codesandbox.io/s/async-pine-gogdb

slapner:
<img width="675" alt="Screen Shot 2020-05-15 at 9 19 01 AM" src="https://user-images.githubusercontent.com/20480013/82054627-37444e80-968d-11ea-8a47-9cdafc72052d.png">

The error is present in both the codesandbox links

syunto07ka:
it's my installed dependences in pakcage.json.
so yeah, "16.13.1" is correct.

```
"dependencies": {
    "@testing-library/jest-dom": "^4.2.4",
    "@testing-library/react": "^9.5.0",
[...]

PR #369, opened 2020-06-20
Do not cause Batcher update while call getRecoilValueAsLoadable

markstock7:
trying to fix https://github.com/facebookexperimental/Recoil/issues/12

before the commit , test will report the following warning.

image

after this commit 

image


mondaychen:
@davidmccabe @drarmstr

desmap:
@davidmccabe @drarmstr When will this be merged and released? Thanks

drarmstr:
Thank you so much for looking into this issue and trying to fix it!  One of the goals of Recoil is to work with React Concurrent Mode.  As part of that effort we are working on a diff that removes the data-flow graph of dependencies out of the  `TreeState`.  This change avoids updating state, and thus causing Batcher updates, during reads, and I think actually obviates the need for this PR.  We are getting closer to being able to land that..

desmap:
@drarmstr thanks for the quick reply and I am looking fwd!

markstock7:
> Thank you so much for looking into this issue and trying to fix it! One of the goals of Recoil is to work with React Concurrent Mode. As part of that effort we are working on a diff that removes the data-flow graph of dependencies out of the `TreeState`. This change avoids updating state, and thus causing Batcher updates, during reads, and I think actually obviates the need for this PR. We are getting closer to being able to land that..

it there any way that I can know the new implementation in advance？

drarmstr:
We're just moving the data-flow graph from `TreeState` to `StoreState`, with versioning.

jonthomp:
@drarmstr would the solution in this MR work in the short-term, or does it have potential issues in itself?

drarmstr:
The diff has landed in master, so closing this now.

mrlubos:
Hi @drarmstr, any update on the ETA for `0.0.11`? Thanks!

PR #383, opened 2020-06-23
getNodes accessor for Snapshot

drarmstr:
Summary:
Add `getNodes` accessor to the `Snapshot` interface

Provide the following interface to allow us to query sets of nodes based on our usage:

```
class Snapshot {
  ...
  getNodes_UNSTABLE: ({
    types?: Array<'atom' | 'selector'>,
    dirty?: boolean,
  }) => Iterable<RecoilValue<mixed>>;
}
```
* **`types`** - An array specifying the types of nodes to return including `atom` and `selector`.  The default is both.
* **`dirty`** - Only return dirty nodes.  Currently this only supports dirty atoms, but we should probably support dirty selectors for debug tools.

To keep scalability in mind, this interface returns an Iterable using a generator function so that we never produces temporary arrays containing all nodes.

NOTE: Feedback on the API for query parameters/options is welcome...  I'm not super keen on the current approach, which by default reports initialized nodes.  The set of initialized nodes in a snapshot my change as selectors are evaluated and discover more.  However, even if we report all globally registered nodes it's really the same problem as nodes can still be dynamically created during selector evaluation.

Reviewed By: csantos42

Differential Revision: D21919550


nikhilag:
@drarmstr Thanks for making this change. It seems this change wasn't made for typescript yet. Should I go ahead and fix it?

drarmstr:
@nikhilag - we have a pending diff to add that, hopefully it will be landing soon.

PR #392, opened 2020-06-23
Disable react/prop-types

jacques-blom:
Disable the `react/prop-types` ESLint rule. It is causing CI to fail because of recently added components in tests. I don't think this rule is needed since we can statically type component props with Flow.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,7 @@ module.exports = {
     strict: 0,
     'jsx-a11y/href-no-hash': 'off',
     'react/jsx-key': 'off',
+    'react/prop-types': 'off',
   },
   settings: {
     react: {


behnammodi:
@jacques-blom Awesome, But why `Facebook Internal - Builds & Tests` is failing

jacques-blom:
You'd have to ask @drarmstr 😊

behnammodi:
@drarmstr It's awesome for helping to test, please check it.

drarmstr:
I don't think there is any real issue, I'm just working on getting it landed.  Thanks for helping with this!

behnammodi:
@drarmstr Also on `master` branch we have some issue like this:

Screen Shot 2020-06-26 at 00 15 28

PR #396, opened 2020-06-24
Fix code error in docs

dkovacevic15:
Added a destructuring to the hook return value.


drarmstr:
[`useSetRecoilState()`](https://recoiljs.org/docs/api-reference/core/useSetRecoilState) returns a setter function directly and does not require a destructuring like [`useRecoilState()`](https://recoiljs.org/docs/api-reference/core/useRecoilState).  You would use this hook when you do not want the component to subscribe to re-render when the state changes.

PR #402, opened 2020-06-25
Add eslint command on lint-staged

behnammodi:
I think, we need `eslint --fix` after `prettier --write` in `lint-staged`


drarmstr:
cc @aaronabramov, @mondaychen

mondaychen:
@aaronabramov is it safe to enable this? I don't think it's enabled by default internally.

aaronabramov:
i think so? i'm not sure what `lint-staged` does actually :) but i think eslint should be very similar to internal and we'll probably see if anything goes worng

PR #419, opened 2020-06-30
Fix word spelling errors

Holybasil:


docs/docs/guides/persistence.md
```diff
@@ -54,7 +54,7 @@ Here is a basic example:
 ```jsx
 const initializeState = ({set}) => {
   for(const [key, value] of Storage.entries()) {
-    set(getAtomWithKey(key), JSON.parse(value)).value;
```
drarmstr, line 57: Given the example code above, it was serialized as an object with a `value` property.

Holybasil: Got it!
```js
Storage.setItem(
  modifiedAtom,
  JSON.stringify({value: atomValues.get(modifiedAtom)}),
);
```
It should be `JSON.parse(value).value` rathar than `set().value`.

PR #413, opened 2020-06-28
feature: Build ESM alongside CommonJS

pocket7878:
Building EMS module alongside CommonJS to use with webpack + babel.

PR #433, opened 2020-07-04
fix: Fix rollup.config.js to handling CommonJS, ESM, ESM for Browser, UMD correctly.

pocket7878:
FIX issue from previous PR: https://github.com/facebookexperimental/Recoil/pull/413


pocket7878:
@mondaychen 
I'v started draft PR to fix issue.

mondaychen:
Can you extract the common parts?

JeremyRH:
I've checked out this branch and inspected the 5 outputs. Here are my notes:

#### dist/recoil.js
- UMD format.
- Development mode (contains `console.error/warn` calls).
- If the environment is using globals, relies on `global.react` and `global.reactDom`. This should probably be changed to `global.React` and `global.ReactDOM` as [those are the globals React and ReactDOM define](https://github.com/facebook/react/blob/baff5cc2f69d30589a5dc65b089e47765437294b/scripts/rollup/bundles.js#L64).
- Used as the default unpkg file.

#### dist/recoil.min.js
- UMD format.
- Production mode (no `console.error/warn` calls).
- Same issue as dist/recoil.js with global `react/reactDom`.
- This file has no "default entry". Using it would require knowing the path: `https://unpkg.com/recoil@0.0.10/dist/recoil.min.js` or `require("recoil/dist/recoil.min.js")`. Not a big problem but changing these paths might be considered a breaking change in the future.

#### es/recoil.js
- ES modules format.
- Development mode.
- Used as the default module file.

#### es/recoil.mjs
- ES modules format.
- Production mode.
- This file has no "default entry". Using it would require knowing the path: `import Recoil from "recoil/es/recoil.mjs"`.

#### lib/recoil.js
- CommonJS format.
- Contains the `process.env.NODE_ENV !== "production"` check before printing development warnings/errors. Most minifiers will remove this so the consumers build process should determine if this is ultimately used in development or production mode.
- Used as the default CommonJS file.

___

I would change the `react` and `reactDom` globals to `React` and `ReactDOM` for the UMD outputs. Use `output.globals` in the rollup config.

The "module" field in package.json points to es/recoil.js. Some (most?) bundlers will use the module entry if the developer uses `import` to import Recoil. That means they will be getting the development mode and `console.error/warn` will be called in their production builds. I don't know of a way to dynamically choose the module entry.
You could just leave the `process.env.NODE_ENV !== "production"` in place for es/recoil.js and anyone trying to use this directly in the browser needs to use es/recoil.mjs.
I don't like that solution but I can't think of a nicer way to handle this right now.

Everything else looks great :+1:

pocket7878:
@JeremyRH Thank you for your review!
I'm just updated rollup.config.js!

mondaychen:
@pocket7878 can you rebase on top of master?

mondaychen:
Hi @pocket7878 

Our sync job was broken since yesterday, unfortunately your PR was partially merged...
Can you rebase on top of master, and create another PR? I believe your changes based on @JeremyRH's suggestion was not included.

PR #434, opened 2020-07-04
chore: update flow-bin version

Naturalclar:
updated flow-bin to 0.129.0

starting from flow-bin v0.127, error suppression syntax has been standardized, and  suppress_comment option for .flowconfig is no longer supported


drarmstr (commented):
Please avoid unnecessary prettier syntax updates.

PR #435, opened 2020-07-05
[docs/example] Update selector.md async example

eLeontev:
current example with async selector is not clear, since without inner await we could remove `async` prefix and directly return callback's result if the result is a promise
```
get: async () => {
    return myDBQueryPromise();
}
```

will be the same
```
get: () => myDBQueryPromise()
```

or to be sure to work with promises always simple wrap the result `get: () => Promise.resolve(myDBQueryPromise())`

to avoid possible misunderstanding I propose to add `await` in the callback's body to highlight `async` root of the get callback

The result:
```
get: async () => {
    const queryResults = await myDBQueryPromise();
    // other dependent operations
    
    return queryResults;
  }
```

### DIFF
--- a/docs/docs/api-reference/core/selector.md
+++ b/docs/docs/api-reference/core/selector.md
@@ -164,7 +164,8 @@ import {selector, useRecoilValue} from 'recoil';
 const myQuery = selector({
   key: 'MyDBQuery',
   get: async () => {
-    return myDBQueryPromise();
+    const response = await fetch(getMyRequestUrl());
+    return response.json();
   },
 });
 


docs/docs/api-reference/core/selector.md
```diff
@@ -164,13 +164,16 @@ import {selector, useRecoilValue} from 'recoil';
 const myQuery = selector({
   key: 'MyDBQuery',
   get: async () => {
-    return myDBQueryPromise();
+    const queryResults = await myDBQueryPromise();
+    // other dependent operations
```
drarmstr, line 168: How about:
```jsx
get: async () => {
  const queryResults = await myDBQueryPromise();
  return queryResults.data;
},
```

eLeontev: sure, in case if you agree to add more prectical content I could replace it to smt like this
```
get: async () => {
  const response= await fetch(getMyRequest());
  return await response.json();
},
```

let me know which way do you prefer. Once the improvement will be confirmed I will update the PR.

drarmstr: haha, maybe `JSON.parse(response)`.  But, seems `getMyRequestUrl()` would really be actually doing the request, not just getting the URL, so `getMyRequestUrl()` doesn't sound right.

eLeontev: `fetch(https://path/to/url.json')` or `fetch(getUrlToFetchJSON())`?

drarmstr: sure, with the `fetch()`: `await fetch(getMyRequestUrl())`

eLeontev: it's what I mentioned initially. Anyway I will apply my example, please will comment what do you suppose to replace (for now I misundersand your point)

drarmstr: I just missed the `fetch()` part, sorry.  But, please update `response.json()` to `JSON.parse(response)` or `JSON.parse(response.date)`

eLeontev: could you clarify how do you expect it will work with any real request (common application/json content-type)?
my check raises correct execption
image

and current version works as expected
image

drarmstr: Ah, I see `fetch()` is a specific Web API that returns an object with `json()`.

PR #437, opened 2020-07-05
[docs] Update useResetRecoilState.md

eLeontev:
looks like `useResetRecoilState` setter call  will not trigger component re-rendering. 
I did not have a chance to validate it, but I sure it's true.

I propose to add the same phrase as in https://recoiljs.org/docs/api-reference/core/useSetRecoilState about re-rendering

In case if I'm wrong please decline the PR and possible it could be the reason to fix the behaviour since from API-usage point both setters  (`useResetRecoilState`, `useSetRecoilState`)are the same.

### DIFF
--- a/docs/docs/api-reference/core/useResetRecoilState.md
+++ b/docs/docs/api-reference/core/useResetRecoilState.md
@@ -5,6 +5,8 @@ sidebar_label: useResetRecoilState()
 
 Returns a function that will reset the value of the given state to its default value.
 
+Using **useResetRecoilState()** allows a component to reset the state to its default value without subscribing to re-render whenever the state changes.
+
 ---
 
 ```jsx


drarmstr:
@eLeontev - `useResetRecoilState()` will cause subscribers of the state to re-render, but will not cause the component using the hook to subscribe to re-render when the state value changes.  Perhaps we should be more explicit about this.

eLeontev:
@drarmstr, sure and looks like it was missed on the initial commit of this doc.

In this case could my update be accepted, since it highlighs this fact in the same manner as in `useSetRecoilState`?

Link to the source with `useSetRecoilState`: https://github.com/facebookexperimental/Recoil/blame/docs/docs/docs/api-reference/core/useSetRecoilState.md#L22

and the screen:
image

docs/docs/api-reference/core/useResetRecoilState.md
```diff
@@ -5,6 +5,8 @@ sidebar_label: useResetRecoilState()
 
 Returns a function that will reset the value of the given state to its default value.
 
+Using **useResetRecoilState()** allows a component to set the default value without re-rendering when the value changes.
```
drarmstr, line 8: "allows a component to reset the state to its default value without subscribing to re-render whenever the state changes."

eLeontev: Done. tThe description was updated based on proposed text.

PR #452, opened 2020-07-08
export flow typings

### LINKED ISSUE #165: Flow support
Recoil is written in Flow, but it's published without the types so the information is lost. Would be great if Flow types were published with the library.

Naturalclar:
fixes #165 
adds flow type definition in built package


Naturalclar:
CI Error is a separate issue which is handled in #450

mondaychen:
Like you mentioned, this plugin requires publishing the whole src folder. I'll try explore other options if possible.

PR #467, opened 2020-07-11
Transpile es5

### LINKED ISSUE #238: ES6 code in production build breaks IE11
I was hoping to give Recoil a try in one of my projects, but it looks like ES6 code is not being transpiled away in the production build:

```js
const next=new Set(set);
class AbstractRecoilValue{constructor(newKey)
```

We sadly have to support IE11, would it be possible to update the rollup config to output ES5 code instead?

Brianzchen:
Fixes #238 
Add `@babel/preset-env` to rollup config which will transpile code down to es5

Does nothing to `new Set()` as that needs to be polyfilled and I haven't tested this on ie11 to see if that's necessary which could come in as a separate PR.


mondaychen:
Can you get rid of the changes in src?

mondaychen:
What's our goal of transpiling down to ES5? Do we want to make it work on IE 11 out-of-box?
If people need to add polyfill for `Set`, can we still say it works out-of box?

Another thing we should consider is which build targets we enable ES5 tranpiling. We have multiple build targets now. I think for ES and CommonJS builds, we expect people to use their bundlers, and it should be easy to add ES5 transpiling in their config file. For UMD builds, people can potentially use them directly in old browsers. For ES for browser build, people can only directly use it in modern browsers.

Brianzchen:
> What's our goal of transpiling down to ES5? Do we want to make it work on IE 11 out-of-box?
> If people need to add polyfill for `Set`, can we still say it works out-of box?
> 
> Another thing we should consider is which build targets we enable ES5 tranpiling. We have multiple build targets now. I think for ES and CommonJS builds, we expect people to use their bundlers, and it should be easy to add ES5 transpiling in their config file. For UMD builds, people can potentially use them directly in old browsers. For ES for browser build, people can only directly use it in modern browsers.

Yes the goal is to ship IE11 out of the box. I actually intended to make an incremental change but I've got my hands on a windows laptop from work. So I'll work on this further to make it runnable on IE11.

I think besides `es` the rest should be runnable out of the box in IE11. Since as I mentioned in one of the issues, typically a bundler doesn't transpile third party deps and only rolls them into the bundle with the setting, `exclude: /node_modules/`.

JeremyRH:
@mondaychen @Brianzchen Webpack will use the "main" output by default but can also use the "module" output. The output `es/recoil.js` also needs to be transpiled to ES5.

Brianzchen:
> @mondaychen @Brianzchen Webpack will use the "main" output by default but can also use the "module" output. The output `es/recoil.js` also needs to be transpiled to ES5.

That's right, I've just run some tests and need to work on my solution more. But webpack will pull the `es` module and will fail in edge and ie11. One of the issues is it can't parse the spread operator

anajavi:
> If people need to add polyfill for `Set`, can we still say it works out-of box?

React itself seems to require Set and Map polyfills: https://reactjs.org/docs/javascript-environment-requirements.html

ajaybeniwal:
Is there any update on this will it be available for consumption in ie11

Brianzchen:
@ajaybeniwal Hey sorry, I've been busy with other projects so haven't done much work on this. My current solution doesn't actually work right now so I'd probably reevaluate the whole thing.

Given what I have on my plate, I'd work on this maybe in a few weeks, so if someone else can pick it up that'd be awesome.

mondaychen:
@ajaybeniwal if you are using webpack or babel, you can add `@babel/preset-env` to your own project and transpile recoil to IE11 compatible code.

mondaychen:
A Recoil team member expressed concerns on performance impact once we introduce ES5 transpiling. I hope we can built some simple performance benchmarking so we can feel comfortable shipping it.

mondaychen:
A note for @Brianzchen or someone who plan to pick this up: let's enable ES5 transpiling on UMD builds first. When we release 0.0.11, I will include a paragraph to explain how to use babel's preset-env for those who install it via NPM (I assume most people are using bundlers so it shouldn't be too hard to add that). In the future, we'll study the impact on performance and then decide if we do the same for all the builds.

Also, it would be very helpful if you can include the difference in output files (with and without ES5).

Thanks!

Brianzchen:
@mondaychen can you give some more context on how there's performance issue? Just that it's extremely rare for a web library to not ship with es5/ie11 compatibility out of the box and would make recoil quite non-standard in a way.

mondaychen:
Some concerns on performance were raised in issues like #228. When it comes ES5, I think the major concern is related to data structures we use like `Map` and `Set`. Currently we do a lot of copying on these data structures, which has nontrival impact on performance if you have a lot of atoms. Some ES5 polyfills might make it worse.
cc @davidmccabe

PR #468, opened 2020-07-11
Expose flow typings with gen-flow-files

### LINKED ISSUE #165: Flow support

Brianzchen:
Fixes #165 
Use https://github.com/lessmess-dev/gen-flow-files to generate flow type definitions for consuming applications

This generates tons of flow errors that can't be resolved though so users have to add the dependency to their `[declarations]`

```
[declarations]
.*node_modules/recoil/.*
```

Otherwise you get errors such as:
Screen Shot 2020-07-11 at 7 47 03 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
   "license": "MIT",
   "scripts": {
     "prepare": "install-peers",
-    "build": "rollup -c && node scripts/postbuild.js",
+    "build:flow": "gen-flow-files src --out-dir lib && mv lib/Recoil_index.js.flow lib/recoil.js.flow",
+    "build": "rollup -c && node scripts/postbuild.js && yarn build:flow",
     "test": "jest src/*",
     "format": "prettier --write \"./**/*.{js,md,json}\"",
     "flow": "flow --show-all-errors",
@@ -52,6 +53,7 @@
     "eslint-plugin-jest": "^23.13.2",
     "eslint-plugin-react": "^7.20.0",
     "flow-bin": "^0.126.1",
+    "gen-flow-files": "^0.4.11",
     "husky": ">=4",
     "immutable": "^4.0.0-rc.12",
     "install-peers-cli": "^2.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -8,6 +8,13 @@
   dependencies:
     "@babel/highlight" "^7.10.1"
 
+"@babel/code-frame@^7.10.4":
+  version "7.10.4"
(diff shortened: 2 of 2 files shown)


Brianzchen:
@mondaychen This uses gen-flow-files as you recommended.

Tested and it works but I also have types-first enabled so not sure what I should do when I export my `atom`, probably needs more work later after the initial exposure, currently having to export as type `any`

```
// @flow
import { atom } from 'recoil';

type State = {
  text: string,
  type: NotificationType,
};

export const notificationState = (atom<State>({
  key: 'notificationState',
  default: {
    text: '',
    type: 'success',
  },
}): any);
```

mondaychen:
Thanks! I should be able to get back to you on Tuesday

mondaychen:
OK I think we are good with this approach. 

You mentioned "This generates tons of flow errors that can't be resolved though so users have to add the dependency to their `[declarations]`". Is there something we can do to make this easier?

Brianzchen:
I've done further testing.

- My initial analysis used `yarn link` which throws errors probably caused by the `Recoil` project itself.
- But when I just copy the flow files over to my application project inside `/node_modules/recoil`, it throws no errors when running `yarn flow` and I am able to get flow type analysis in my application.
- If I were to instead put these files in my `src` dir `yarn flow` will complain

But if I open files in node_modules there's clearly flow errors coming from my IDE though with no impact to usage,
Screen Shot 2020-07-17 at 9 02 20 am

Overall seems like this wouldn't have an impact on users and flow somehow understands that theses are declaration files from a third party and knows not to throw errors in flow analysis?

Brianzchen:
@mondaychen anymore work needs to go into this?

Brianzchen:
@drarmstr I think this PR is complete, don't think there's any more work needed, can you advise?

mondaychen:
Sorry @Brianzchen for being slow on this. I will create a flow-based application later this week and see how well it works. Then I'll get back to you.

sherwinyu:
Hello, I'm trying to get flow types working for React but I still encounter `Cannot resolve module recoil.` when typechecking with `flow`. I'm using recoil v0.4.1.

Any guidance on how to actually get flow to pick up these types? Do I need to do anything else with .flowconfig?

Brianzchen:
> Any guidance on how to actually get flow to pick up these types? Do I need to do anything else with .flowconfig?

Looks like recoil still ships with flow types. What does your flowconfig look like? Best to make sure you're not excluding recoil in your ignore group.

Issue #208, opened 2020-05-30
Export useRecoilInterface()


natew:
Writing a library around recoil and having this would save a lot of overhead.

drarmstr:
We're hoping to deprecate this API.  What aspect of it do you find useful so we can consider requirements for alternative mechanisms to support it?  (e.g. #134)

alidlorenzo:
@drarmstr after glancing at code, I was also looking to use `useRecoilInterface` so that I could build my own state management utilities on top of recoil, without calling multiple `useRecoilState` hooks.

from reading #134, it seems like the recommended approach is to use `useRecoilCallback` .

drarmstr:
Yes, it is recommended to use a supported mechanism such as `useRecoilCallback()`.  Note that some ongoing work we are doing to enable React Concurrent Mode is changing the semantics of `useRecoilInterface()`, and we expect to deprecate it.

natew:
Yea, I re-opened a different issue here, I forgot I hadn't gotten back here!

Can see example of `useRecoilCallback`.

https://github.com/facebookexperimental/Recoil/issues/451#issuecomment-655243901

PR #474, opened 2020-07-13
expose useRecoilInterface

alidlorenzo:
I'd be great to have access to `useRecoilInterface` so that we can build more traditional global store management facilities on top of Recoil


alidlorenzo:
ah did not see #208, will go ahead and close this and discuss in issue

drarmstr:
Duplicate of #208

PR #483, opened 2020-07-17
Documentation for 0.0.11 [PENDING]

drarmstr:
This PR is collecting ongoing pending documentation for 0.0.11 for when it is ready to release.

* Concurrent Mode
* Performance
* Types & Packaging
* Context Bridging
* Snapshot dev tools API
* ...


mlajtos:
How would you use `useRecoilBridgeAcrossReactRoots_UNSTABLE` hook with `ReactDOM.render`? Like this?

```js
const Bridge = ({ children }) => {
    const RecoilBridge = useRecoilBridgeAcrossReactRoots_UNSTABLE();
    return <RecoilBridge>{children}</RecoilBridge>;
};

ReactDOM.render(
    <Bridge><ComponentUsingRecoilStateFromAnotherRoot /></Bridge>,
    document.querySelector("#someReactRoot")
);
```

drarmstr:
@mlajtos 
> How would you use `useRecoilBridgeAcrossReactRoots_UNSTABLE` hook with `ReactDOM.render`? Like this?
> 
> ```js
> const Bridge = ({ children }) => {
>     const RecoilBridge = useRecoilBridgeAcrossReactRoots_UNSTABLE();
>     return <RecoilBridge>{children}</RecoilBridge>;
> };
> 
> ReactDOM.render(
>     <Bridge><ComponentUsingRecoilStateFromAnotherRoot /></Bridge>,
>     document.querySelector("#someReactRoot")
> );
> ```

There is an example in the tests in [this commit](https://github.com/facebookexperimental/Recoil/commit/d278015084b940e083205e51ae070cbc3747d190).  The hook needs to be called in the context of the parent host React and then used in the nested React root:

```jsx
function NestedReactRoot({children}) {
  const ref = useRef();
  const RecoilBridge = useRecoilBridgeAcrossReactRoots();

  useEffect(
    () =>
      act(() => {
        ReactDOM.render(<RecoilBridge>{children}</RecoilBridge>, ref.current);
      }),
    [children],
  );
  return <div ref={ref} />;
}

function MyApp() {
  return (
    <RecoilRoot>
      ...
      <NestedReactRoot>
        ...
      </NestedReactRoot>
    </RecoilRoot>
  );
}
```

mlajtos:
That makes sense. Thank you 🙂

yomybaby:
@drarmstr Thank you for your effort.
After looking your PR, I have a question. 
If not nested like below, there is no way to bridge now. Isn't it?
```
ReactDOM.render(<App />, document.getElementById('root'));
ReactDOM.render(<App2 />, document.getElementById('root2'));
```
As react-native-navigation screens are not part of the same component tree(not nested). How can I share recoil storeRef between different component tree?
https://wix.github.io/react-native-navigation/docs/third-party-react-context

Issue #490, opened 2020-07-20
[request] add new entity only with set functionality


eLeontev:
I faced with the required code in selector to implement common setter only to trigger other states.
**Business case:**
I have multiple places where I need to cleanup my form values and then reset my state to not dirty.
``` javascript
const textState = atom({key: '1', default: ''});
const dropdownState = atom({key: '2', default: 'valueId'});

const dirtyState = atom({key: '3', default: false});

const resetState = selector({
    key: '4',
    get: () => {}, // redundant but required,
    set: ({set}) => {
        set(textState, '');
        set(dropdownState, 'valueId');
        set(dirtyState, false);
    }
})
```  

and in the components where the logic must be called:
```javascript
const Component = () => {
    const resetState = useSetRecoilStat(resetState);
    return <button onClick={resetState}>click me</button>; 
}
```

**Request:** add new entity with opportunity to use only setter like initial trigger of state reactions.

```javascript
const resetStateSetter = power<TypeOfValueToSet>({
    key: '4', // maybe is redundant for public API
    set: ({set}, value: TypeOfValueToSet) => {
        set(textState, '');
        set(dropdownState, 'valueId');
        set(dirtyState, false);
    }
})
```
I named it after quick look as `power` like smt abstract affects/triggers atoms (temporary name)

eLeontev:
proposal: #491

drarmstr:
Do you really need a selector here?  How about just making a hook:

```jsx
const useResetState = () => {
  const setTextState = useSetRecoilState(textState);
  const setDropdownState = useSetRecoilState(dropdownState);
  const setDirtyState = useSetRecoilState(dirtyState);
  return () => {
    setTextState('');
    setDropdownState9('valudId');
    setDirtyState(false);
  };
};
```
or, `useRecoilCallback()`:
```jsx
const resetState = useRecoilCallback(({set}) => () => {
  set(textState, '');
  set(dropdownState, 'valueId');
  set(dirtyState, false);
}, []);
```

eLeontev:
Looks fine, ony one point I want to highlight: this approach is placed under React more than Recoil and on my point is not consistentdoes not look obvious. 

I mean we have default chain of actions are fired step by step started with initial signal (any setter) and in this case we have two different ways for the same result.

e.g. want to initiate state changes
 do we have values (state) in initial setter 
   ? let directly use setter with updated value
   : let perfrom smt other like castom hook or `useRecoilCallback`

[...]

PR #491, opened 2020-07-20
#490 proposal: add the new entity setter only

eLeontev:
proposal PR for request #490


drarmstr:
See discussion in #490 - Selectors represent derived state.  If we are only updating state then we can use hooks or Recoil callback mechanisms to update state and don't need the overhead of managing a selector node in the data-flow graph for downstream derived state.

eLeontev:
during the descussion was decided to use react hooks or `useRecoilCallback` as correct way for initiator

PR #500, opened 2020-07-27
Don't use fake timers in tests

aaronabramov:
fake timers are a pain to debug and work with. These tests are currently breaking, because internally at fb promises work differently and it's pretty painful to debug why and how. I removed all fake timers in this tests and added a `waitFor` utility, that will wait for the entire thing to happen naturally with real timers (as it would in the real world).

cons of this approach:
- since this is using real timers, it actually has to wait for things to happen. It will result in slower tests, but practically it's only a 10ms difference. 
- some cases need additional infra to be built in order to test these cases. E.g. we need to have hooks to manually resolve some selectors (instead of relying on X number of ticks for them to be resolved). If we actually build this infra (i mean it's like 10 lines of code) the test will also become much more maintainable and readable, since there' will not be obscure flows depending on timer ticks

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,8 @@
       "src/hooks/__tests__/Recoil_PublicHooks-test.js",
       "src/hooks/__tests__/Recoil_useRecoilCallback-test.js",
       "src/recoil_values/__tests__/Recoil_atomFamily-test.js",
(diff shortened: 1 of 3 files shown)


aaronabramov:
i don't think this will work

PR #540, opened 2020-08-29
add installation guide on ES5 & CDN

mondaychen:


drarmstr:
@mondaychen, I inadvertently made a copy of this branch here in the `facebookexperimental/Recoil` repository.  We can't merge into it on GitHub since only the `docs` branch allows direct merging.  Could you PR or merge this commit into `drarmstr/docs_0_0_11` so then we can have a single atomic PR to `facebookexperimental/Recoil/docs` branch with the release?

PR #554, opened 2020-09-04
Class Snapshot Text

ricardoham:
Should the correct text of this section be as follows?

 [...] It is intended to standardize the API for observing, inspecting, and managing global Recoil state.  It is mostly useful for dev tools, global state synchronization, history navigation, etc.


drarmstr:
`&` is a ligature for `et`, these are equivalent.  Not as many people are familiar with it, it seems, so fine to update.

PR #555, opened 2020-09-05
fix: pass CI tests

mallchel:
console error prevents the success of GitHub actions workflow. This PR is just setting up running CI tests, mocking console.error and console.warn


mallchel:
@aaronabramov 
console error prevents success github actions workflow. This PR is just setting up running CI tests, mocking console.error and console.warn

drarmstr:
cc @aaronabramov

mallchel:
@drarmstr it seems like tests were failing only in this line ```expect(c.textContent).toBe('"ATOMATOMATOM"');```
We can remove mocks for console.error and console.warn for CI

drarmstr:
> @drarmstr it seems like tests were failing only in this line `expect(c.textContent).toBe('"ATOMATOMATOM"');`
> We can remove mocks for console.error and console.warn for CI

If tests are failing there, could it be an actual failure or test we want to fix?  This PR seems dangerous to hide problems?

mallchel:
@drarmstr No, this PR doesn't hide failed tests. If some test fails with the "expect" expression, then the exit code would be non zero and GitHub's actions fail.

Mocking for console methods doesn't make sense at the moment. I can just remove it together with the test-ci script. What do you think?

drarmstr:
I see.  I also fixed the `act()` issue with our local version of #561 (not currently visible on GitHub, the public PR isn't valid).  Anyway, I'll leave this for @aaronabramov to update when he returns next week.

mallchel:
@aaronabramov I'm not sure how to properly test the useRecoilValueLoadable

src/recoil_values/__tests__/Recoil_selector-test.js
```diff
@@ -768,8 +768,8 @@ describe('Async selector resolution notifies all stores that read pending', () =
     expect(c.textContent).toEqual('loading');
 
     act(() => resolve('bar'));
-    await flushPromisesAndTimers();
-    await flushPromisesAndTimers();
+    await act(flushPromisesAndTimers);
```
drarmstr, line 771: These are no longer necessary with #576

drarmstr (changes requested):

src/recoil_values/__tests__/Recoil_selector-test.js
mallchel: @drarmstr Yeah. I missed it. fixed

src/util/Recoil_mutableSource.js
```diff
@@ -0,0 +1,12 @@
+const React = require('React');
+
+function mutableSourceIsExist() {
```
aaronabramov, line 3:
```js
function mutableSourceExists() {
```

mallchel:
@aaronabramov updated a PR, moved useMutableSource to Recoil_mutableSource.

PR #557, opened 2020-09-06
Coding style corrections

suren-atoyan:
Hello guys. I've just started to investigate Recoil. Thank you for your work; I really love it.

During reading Recoil docs I found some incompatibilities of the coding style of some examples. So, I didn't add new rules (which I would like to do), I just made them look similar; in some code segments, you forgot semicolons and trailing commas. Unfortunately, the process of checking was manual.


drarmstr:
Thanks @suren-atoyan !

PR #561, opened 2020-09-08
test(getSubscriptions): Fixed warning message for getSubscriptions

benhalverson:
I saw this warning in the tests so I've fixed it.
```console.error
      Warning: An update to Batcher inside a test was not wrapped in act(...).

      When testing, code that causes React state updates should be wrapped into act(...):

      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */

      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
          in Batcher (created by RecoilRoot)
          in RecoilRoot```

### DIFF
--- a/src/hooks/__tests__/Recoil_useRecoilSnapshot-test.js
+++ b/src/hooks/__tests__/Recoil_useRecoilSnapshot-test.js
@@ -155,65 +155,66 @@ test('useRecoilSnapshot - async selectors', async () => {
   expect(snapshots[0].getLoadable(mySelector).contents).toEqual('RESOLVE');
 });
 
-test('getSubscriptions', async () => {
-  const myAtom = atom<string>({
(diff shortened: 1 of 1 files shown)


drarmstr (changes requested):
It seems like this warning only shows up with the OSS environment and not our internal environment..  

But, anyway, unfortunately this solution doesn't work because the `except()` checks don't actually flag failures when they are used in the `act()` callback.

drarmstr:
> It seems like this warning only shows up with the OSS environment and not our internal environment..
> 
> But, anyway, unfortunately this solution doesn't work because the `except()` checks don't actually flag failures when they are used in the `act()` callback.

Ok, seems like it's sufficient if we wrap the `await flushPromisesAndTimers();` calls with `act()`

PR #575, opened 2020-09-15
Fix TypeScript types: Couldn't pass ReadonlyArray as SerializableParam.

hachibeeDI:
Previously, we cannot pass ReadonlyArray as a second type argument of `selectorFamily`:

```typescript
const sel = selectorFamily<bool, ReadonlyArray<number>>(  //...
```

After fix, the code above is going to be valid and we can still pass a normal `Array` as the type argument.

------

I confirm `yarn test:typescript` passed

```typescript
$ yarn test:typescript
yarn run v1.22.5
$ dtslint typescript
✨  Done in 5.65s.
yarn test:typescript  8.75s user 0.46s system 153% cpu 6.015 total
```

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -205,7 +205,7 @@ export function isRecoilValue(val: unknown): val is RecoilValue<any>; // eslint-
 // bigint not supported yet
 type Primitive = undefined | null | boolean | number | symbol | string;
 
-export type SerializableParam = Primitive | SerializableParam[] | { [key: string]: SerializableParam };
+export type SerializableParam = Primitive | ReadonlyArray<SerializableParam> | { [key: string]: SerializableParam };
 
 export interface AtomFamilyOptions<T, P extends SerializableParam> {
   key: NodeKey;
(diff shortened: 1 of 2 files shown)


typescript/index.d.ts
```diff
@@ -205,7 +205,7 @@ export function isRecoilValue(val: unknown): val is RecoilValue<any>; // eslint-
 // bigint not supported yet
 type Primitive = undefined | null | boolean | number | symbol | string;
 
-export type SerializableParam = Primitive | SerializableParam[] | { [key: string]: SerializableParam };
+export type SerializableParam = Primitive | ReadonlyArray<SerializableParam> | { [key: string]: SerializableParam };
```
drarmstr, line 208: Read only object as well?

drarmstr (approved):

typescript/index.d.ts
hachibeeDI: I think it's okay as is. That can receive any objects which has readonly properties. 👌

PR #581, opened 2020-09-16
Changelog and package for 0.0.13

davidmccabe:


santinucera:
@davidmccabe When you have in mind add the React Native config for Recoil?

PR #588, opened 2020-09-17
Disable a failing test in OSS build

aaronabramov:
there's some weird stuff going on with number of calls in www and oss builds. there not the same and jumping in a starnge order. This test will still run internally, but i'm disabling it in OSS


src/hooks/__tests__/Recoil_PublicHooks-test.js
```diff
@@ -1407,7 +1410,7 @@ test('Can use an already-resolved promise', async () => {
   expect(container.textContent).toEqual('2');
 });
 
-test('Resolution of suspense causes render just once', async () => {
+fbOnlyTest('Resolution of suspense causes render just once', async () => {
```
drarmstr, line 1413: Add a comment why this is fbonly.

drarmstr (changes requested):

PR #590, opened 2020-09-18
fix: Resolution of suspense causes render just once

mallchel:
Different count of renders is only in components that use useRecoilValue. Because useRecoilValue has multiple behaviors.


drarmstr:
cc @aaronabramov

aaronabramov:
@mallchel awesome! let me try building it internally

PR #591, opened 2020-09-18
Cleanup Handlers for Atom Effects

drarmstr:
Summary:
Add support for Atom Effects to register cleanup handlers as per [this RFC](https://fb.quip.com/bauvAeBEX5Ee).  These handlers will be called when the dynamic atom goes out of scope.  Currently the atoms only go in or out of scope when they are first used by a `<RecoilRoot>` and then when that `<RecoilRoot>` is unmounted.

This is important for cleaning up subscriptions.

## Atom Effects Interface
```
type AtomEffect<T> = ({
  node: RecoilState<T>,
  trigger: 'get' | 'set',

  // Can call either synchronously to initialize value or async to change it later
  setSelf: (
    | T
    | DefaultValue
    | Promise<T | DefaultValue>
    | (T | DefaultValue => T | DefaultValue)
  ) => void,
  resetSelf: () => void,

  // Subscribe to events
  // Called when React batch ends, but before global RecoilTransactionObserver
  onSet: (
    (newValue: T | DefaultValue, oldValue: T | DefaultValue) => void,
  ) => void,
}) => void | () => void; // Return optional cleanup handler
```

## Example
```
const myAtom = atom({
  key: 'MyAtom',
  default: undefined,

  effects: [({node, setSelf, onSet}) => {
    const storage = getStorage(node.key);

    // set a value for synchronous init that overrides default
    setSelf(storage.get());

    // Subscribe to storage updates
    storage.subscribe(value => setSelf(value));

    // Subscribe local changes to write-through to storage
    onSet(value =>
      value instanceof DefaultValue
        ? storage.reset()
        : storage.set(value)
    );

    // Cleanup subscription
    return () => storage.unsubscribe();
  }],
});
```

Differential Revision: D23723600

### DIFF
--- a/src/core/Recoil_FunctionalCore.js
+++ b/src/core/Recoil_FunctionalCore.js
@@ -92,6 +92,11 @@ function setNodeValue<T>(
   return node.set(store, state, newValue);
 }
 
+function cleanupNode(store: Store, key: NodeKey) {
+  const node = getNode(key);
+  node.cleanup(store);
+}
+
 // Find all of the recursively dependent nodes
 function getDownstreamNodes(
   store: Store,
@@ -119,5 +124,6 @@ module.exports = {
   peekNodeLoadable,
   setNodeValue,
   setUnvalidatedAtomValue,
+  cleanupNode,
   getDownstreamNodes,
 };
--- a/src/core/Recoil_Node.js
+++ b/src/core/Recoil_Node.js
@@ -45,6 +45,9 @@ export type ReadOnlyNodeOptions<T> = $ReadOnly<{
   // Returns the discovered deps and the loadable value of the node
   get: (Store, TreeState) => [DependencyMap, Loadable<T>],
 
+  // Cleanup the node when it is removed from a <RecoilRoot>
+  cleanup: Store => void,
+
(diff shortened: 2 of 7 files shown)


mdlavin:
I don't have access to view the RFC, but I was looking for something like this and I have two questions:

1. Will it be possible to use cleanup effects on selectors?  Looking at the code changes, it seems like maybe not. I think it could be very valuable.
2. Assuming cleanup on selectors is allowed, I'd also love to access to dependent recoil values in the cleanup handler.

The motivating use case is very close the one you describe above. The difference would be to replace:
`const storage = getStorage(node.key);` with `get(storageState)`

drarmstr:
@mdlavin - Currently Atom Effects are only for Atoms and not Selectors.  That's because the effects API is intended to support state synchronization and that includes initializing state for first read, selectors set their state by writing to other atoms, and it is more dangerous to allow arbitrary sets synchronously during a get, particularly with Concurrent Mode. 

We do have another outstanding RFC for selector finalization, that may be useful for things like async request cancelation.  But, that also depends on us adding the ability to track when atoms/selectors may go in/our of scope based on any active components subscribing to them.

PR #601, opened 2020-09-22
Fix incorrect types in docs

JustDs:


drarmstr:
An important fix, thanks!

PR #620, opened 2020-09-27
feat: added tests for concurrent mode

mallchel:
Added run same tests in concurrent mode


src/testing/Recoil_CreateReactRoot.js
```diff
@@ -0,0 +1,22 @@
...
+ * @emails oncall+recoil
+ * @flow strict-local
+ * @format
+ */
+
+const React = require('React');
+const ReactDOM = require('ReactDOM');
+
+function createReactRoot(container: HTMLElement, contents: React.Node) {
+  // To test in Legacy Mode
```
aaronabramov, line 16: i don't think it can be called "Legacy" yet 😅
let's call it "Standard (non-concurrent) mode"

jest_configs/concurrent/jest.config.concurrent.js
```diff
@@ -0,0 +1,20 @@
+const path = require('path');
+
+const commonConfig = require('../common.config');
+const currentDir = relative => path.resolve(__dirname, relative);
+
+module.exports = {
+  ...commonConfig,
+  setupFiles: [].concat(commonConfig.setupFiles, [
```
aaronabramov, line 8:
```
[...commonConfig.setupFiles, currentDir('setupJestMock.js')]
```

package.json
```diff
@@ -27,7 +27,7 @@
     "deploy-nightly": "yarn build && node scripts/deploy_nightly_build.js"
   },
   "peerDependencies": {
-    "react": "^16.13.1"
+    "react": "^0.0.0-experimental-94c0244ba"
```
aaronabramov, line 30: i just realized that it'll probably change it for the outside world as well?
@davidmccabe @drarmstr testing this thing in concurrent mode requires the experimental build of react, is that what we run in WWW? 
i'm wondering if there's an easy way to patch the version in test environment

aaronabramov (commented):
@mallchel do you mind adding some description and a screenshot of how test run looks with 2 projects enabled?

src/testing/Recoil_CreateReactRoot.js
mallchel: Oo, I took this term from the official react doc)

jest_configs/concurrent/jest.config.concurrent.js
mallchel: done

aaronabramov:
yeah. sorry. the repo is not ready for this. we'll need to revisit later

PR #628, opened 2020-09-30
feat: added @babel/plugin-transform-modules-commonjs, fix: tests

mallchel:
- feat: added @babel/plugin-transform-modules-commonjs
- fix: failed tests


aaronabramov:
@mondaychen i'm pretty far away from modern JS infra. is that still what we're supposed to do with import/export? they're not in the official spec yet, right?

mallchel:
would be better to reuse one babel config in rollup and jest?

mondaychen:
@mallchel I agree we should use one config for rollup and jest if possible.
Can you provide more information on why we need `@babel/plugin-transform-modules-commonjs`?

@aaronabramov we cannot use import/export yet because it will break `jest.mock` https://github.com/facebook/jest/issues/9430

PR #633, opened 2020-10-04
fix: add check for window  to shouldNotBeFrozen function

RinGoku:
Recoil version is 0.0.13.

## Steps To Reproduce
Run below code and click "Window"  button, then error will be thrown.
https://codesandbox.io/s/weathered-dawn-5xfhh?file=/src/App.js
```
const testState = atom({
  key: "test",
  default: null
});

const Inner = () => {
  const [test, setTest] = useRecoilState(testState);
  return (
    <div>
      value: {test && test.hoge}
      <button
        onClick={() => {
          setTest({ hoge: "fuga" });
        }}
      >
        String
      </button>
      <button
        onClick={() => {
          setTest({
            hoge: window
          });
        }}
      >
        Window
      </button>
    </div>
  );
};

export default function App() {
  return (
    <RecoilRoot>
      <div className="App">
        <Inner />
      </div>
    </RecoilRoot>
  );
}
```
error detail is below.

```
TypeError
Cannot freeze
deepFreezeValue
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:1675:10
deepFreezeValue
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:1680:9
Object.mySet [as set]
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:2143:9
setNodeValue
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:375:15
eval
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:580:30
Object.replaceState
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:1070:18
eval
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:578:60
Object.trace
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:515:10
setRecoilValue
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:578:18
eval
https://8r9gf.csb.app/node_modules/recoil/es/recoil.js:1360:5
onClick
/src/App.js:24:10
```

## Description
This error seems to relate to below line.
https://github.com/facebookexperimental/Recoil/blob/d6aca30332dcf46eaf64b417c3b4715a1678cbb4/src/util/Recoil_deepFreezeValue.js#L70

Object.freeze cannot freeze Window.
So, this PR add a check for Window to the shouldNotBeFrozen function.
(It may be rare to code like above, but the object created by a library I used in my app had the above problem, so I create a PR)

### DIFF
--- a/src/util/Recoil_deepFreezeValue.js
+++ b/src/util/Recoil_deepFreezeValue.js
@@ -56,7 +56,7 @@ function shouldNotBeFrozen(value: mixed): boolean {
     return true;
   }
 
-  if (!!value && value === value.window) {
+  if (typeof window !== 'undefined' && value instanceof window) {
     return true;
   }
 
--- a/src/util/Recoil_deepFreezeValue.js
+++ b/src/util/Recoil_deepFreezeValue.js
@@ -56,6 +56,10 @@ function shouldNotBeFrozen(value: mixed): boolean {
     return true;
   }
 
+  if (!!value && value === value.window) {
+    return true;
+  }
+
   return false;
 }
 
--- a/src/util/__tests__/Recoil_deepFreezeValue-test.js
+++ b/src/util/__tests__/Recoil_deepFreezeValue-test.js
@@ -28,5 +28,8 @@ describe('deepFreezeValue', () => {
       deepFreezeValue({test: new DataView(new ArrayBuffer(16), 0)}),
     ).not.toThrow();
   });
+  test('check no error: object with Window property', () => {
+    expect(() => deepFreezeValue({test: window})).not.toThrow();
+  });
   // TODO add test of other pattern
 });
--- a/src/util/__tests__/Recoil_deepFreezeValue-test.js
+++ b/src/util/__tests__/Recoil_deepFreezeValue-test.js
@@ -29,6 +29,9 @@ describe('deepFreezeValue', () => {
     ).not.toThrow();
   });
   test('check no error: object with Window property', () => {
+    if (typeof window === 'undefined') {
+      return;
+    }
     expect(() => deepFreezeValue({test: window})).not.toThrow();
   });
   // TODO add test of other pattern


src/util/Recoil_deepFreezeValue.js
```diff
@@ -56,6 +56,10 @@ function shouldNotBeFrozen(value: mixed): boolean {
     return true;
   }
 
+  if (!!value && value === value.window) {
```
drarmstr, line 59: Why is it checking for equality of a property called `window`?  Seems like it should check for equality with the global `window`.  Though, for SSR compatibility, it would need some `typeof window !== 'undefined'` check which has been added to the codebase in a few locations.  We should probably normalize that out at this point, especially if it would be used in a loop here.

src/util/__tests__/Recoil_deepFreezeValue-test.js
```diff
@@ -28,5 +28,8 @@ describe('deepFreezeValue', () => {
       deepFreezeValue({test: new DataView(new ArrayBuffer(16), 0)}),
     ).not.toThrow();
   });
+  test('check no error: object with Window property', () => {
+    expect(() => deepFreezeValue({test: window})).not.toThrow();
```
drarmstr, line 32: This also can't access the global `window` directly for SSR compatibility.  We hope to be able to run the test suite in that environment as well one day.  So, need to guard against it with the check described above.

drarmstr (changes requested):

src/util/Recoil_deepFreezeValue.js
RinGoku: @drarmstr 
>Why is it checking for equality of a property called window?
sorry, this is hard to understand.
I fix it to use `typeof window !== undefined && value instanceof Window` instead of above.

src/util/Recoil_deepFreezeValue.js
```diff
@@ -56,6 +56,10 @@ function shouldNotBeFrozen(value: mixed): boolean {
     return true;
   }
 
+  if (typeof window !== 'undefined' && value instanceof window) {
```
drarmstr, line 59: This should check for instance of `Window` vs `window`.  I can fix that and also normalize the `typeof` check.

drarmstr: Ok, I have a fix for the internal diff; though note it won't back-sync to this PR.

PR #643, opened 2020-10-10
fix: CI checks

Naturalclar:
Flow TypeCheck, Jest Testcases, and Lint were failing on the CI.
I've made some fixes to have the ci pass


src/recoil_values/Recoil_selector_NEW.js
```diff
@@ -840,7 +840,6 @@ function selector<T>(
       const message = `Recoil selector has circular dependencies: ${dependencyStack
         .slice(dependencyStack.indexOf(key))
         .join(' \u2192 ')}`;
-      // eslint-disable-next-line fb-www/no-new-error
```
Naturalclar, line 843: this lint rule did not exist for the oss version of recoil, which was causing lint error

src/util/Recoil_useComponentName.js
```diff
@@ -15,7 +15,7 @@ import {useRef} from 'React';
 
 import stackTraceParser from '../util/Recoil_stackTraceParser';
 
-export default function useComponentName(): string {
+module.exports = function useComponentName(): string {
```
Naturalclar, line 18: using export default here caused FlowType inference of this module to be

```
{
  exports: () => string
}
```

instead of just

```
() => string
```

I'm not aware of a way to keep `export default` while retaining the correct type

babel.config.js
```diff
@@ -0,0 +1,34 @@
...
+    '@babel/plugin-proposal-class-properties',
+    '@babel/proposal-nullish-coalescing-operator',
+    '@babel/proposal-optional-chaining',
+    '@babel/transform-flow-strip-types',
+  ];
+
+  if (isTest) {
+    // jest will need import to be converted to require statements.
+    // For release, rollup will transform modules accordingly
+    plugins.push('@babel/plugin-transform-modules-commonjs');
```
Naturalclar, line 27: reason for adding babel-transform-modules-commonjs

scripts/deploy_nightly_build.js
```diff
@@ -1,13 +1,13 @@
-const package = require('../package.json');
+const {version, files, repository} = require('../package.json');
```
Naturalclar, line 1: `package` was a reserved name which causes error by linter

Naturalclar:
Closing and rebasing due to conflict

PR #660, opened 2020-10-19
[Docs] Fix mis-located inner links in persistence/asynchronous-state-sync

Curzy:
fix to navigate right (before: [Page Not Found](https://recoiljs.org/docs/guides/persistence/asynchronous-state-sync))


Curzy:
I have signed CLA right before 😅

mondaychen:
Thanks!

PR #663, opened 2020-10-20
Replace `fregante/setup-git-token` with `setup-git-user`

fregante:
In the v2 of `actions/checkout`, setting the token is no longer necessary, so I created a new config-less action to just set the git user: https://github.com/fregante/setup-git-user

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,9 +12,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     # config GIT so it can do git push
-    - uses: fregante/setup-git-token@v1
-      with:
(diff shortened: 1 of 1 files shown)


fregante:
I signed the CLA, did it work?

mondaychen:
Thanks @fregante. I think I need the `.netrc` part in setup-git-token because I actually clone the repo again in my script and commit from there. See https://github.com/facebookexperimental/Recoil/blob/master/scripts/deploy_nightly_build.js#L21

Maybe I should update my script so I don't need to clone into another folder?

fregante:
If the previous `secrets.GITHUB_TOKEN` is the default one, then this is taken care by `action/checkout` since it now defaults to `preserve-credentials: true`

Issue #671, opened 2020-10-23
[React Native]  Can't find variable Window error and possible solution


a-sane:
Updating the value of recoil state on react native throws the following exception: `Can't find variable: Window` if type of value is array or object

Looks like this line:
https://github.com/facebookexperimental/Recoil/blob/b4db14bbf3135778ea237a80b30239ea9c7fb473/src/util/Recoil_deepFreezeValue.js#L62

from https://github.com/facebookexperimental/Recoil/pull/633 causes this error because there is no Window object on react native

Maybe there can be something like this check?

```js
const isReactNative = typeof navigator !== 'undefined' &&  navigator.product === 'ReactNative';

if (!isSSR && !isReactNative && (value === window || value instanceof Window)) { 
```

drarmstr:
Thanks for proposing a fix for this @a-sane!

PR #672, opened 2020-10-23
[ReactNative] Can't find variable Window exception

a-sane:
https://github.com/facebookexperimental/Recoil/issues/671

### DIFF
--- a/src/util/Recoil_deepFreezeValue.js
+++ b/src/util/Recoil_deepFreezeValue.js
@@ -19,6 +19,7 @@ const isNode = require('./Recoil_isNode');
 const isPromise = require('./Recoil_isPromise');
 
 const isSSR = typeof window === 'undefined';
+const isReactNative = typeof navigator !== 'undefined' && navigator.product === 'ReactNative';
 
 function shouldNotBeFrozen(value: mixed): boolean {
   // Primitives and functions:
@@ -59,7 +60,7 @@ function shouldNotBeFrozen(value: mixed): boolean {
   }
 
   // Some environments, just as Jest, don't work with the instanceof check
-  if (!isSSR && (value === window || value instanceof Window)) {
+  if (!isSSR && !isReactNative && (value === window || value instanceof Window)) {
     return true;
   }
 


davidmccabe:
This seems reasonable to me. Mengdi do you know what's going on with the CI?

mondaychen:
Can we test directly against `Window`?

mondaychen:
@davidmccabe CI has been broken for a while, I'll see if I can do a quick fix

victoriafrench:
> Can we test directly against `Window`?

these are the scripts i have used in the past that may help with this.

```js
function getGlobalScope() {
	// browser
	if (typeof window !== 'undefined') {
		return window;
	}
	// webworkers
	if (typeof self !== 'undefined') {
		return self;
	}
	// node & deno
	if (typeof global !== 'undefined') {
		return global;
	}
	// should not be possible
	return {};
}
```

I have to put it in a function because `const globalScope = window || self || global || {};` would fail.

```js
const globalScope = getGlobalScope();

// web workers: will not have window but will have self, DOM will have both self and window in newer versions,
// they would be the same reference
const isWebWorker = Boolean(typeof window === 'undefined' && typeof self !== 'undefined');

// react native: lifted from recoil
const isReactNative = Boolean(globalScope.navigator && globalScope.navigator.product === 'ReactNative');

// node, deno: would have a global and a process, but filter out react native
const isSSR = typeof global !== 'undefined' && typeof globalScope.process !== 'undefined' && !isReactNative;

// browser: the browser dom would have recursive window and a document, would want to filter out node and dino
// in the event that a browser shim was loaded, though I don't think anyone does that anymore in unit tests. Probably
// better to just ignore isSSR and isReactNative because if they have a browser dom shim, they intend to use it?
const isDOM = Boolean(globalScope.window && globalScope.window.window && globalScope.window.document && globalScope.document && !isReactNative && !isSSR);
```

Maybe something in there would help? But I did test this and it did work in node and the browsers.

```js
function shouldNotBeFrozen(value) {
	if (isDOM && (value === globalScope.window || value instanceof globalScope.Window)) {
		return true;
	}
	return false;
}
```

PR #674, opened 2020-10-26
Update atoms.mdx

luciy:
There is a call to settodolist in the modified method


drarmstr:
using `useRecoilState()` instead of `useSetReacoilState()` would introduce a subscription, which shouldn't be necessary in this case since `todoList` is not referenced in the component.

PR #692, opened 2020-10-30
Fix CI on OSS build

Naturalclar:
OSS build CI was failing due to eslint and flowtype.

I've made some fixes to have CI pass for OSS repo.

cc @drarmstr


src/recoil_values/Recoil_selector_NEW.js
```diff
@@ -840,7 +840,6 @@ function selector<T>(
       const message = `Recoil selector has circular dependencies: ${dependencyStack
         .slice(dependencyStack.indexOf(key))
         .join(' \u2192 ')}`;
-      // eslint-disable-next-line fb-www/no-new-error
```
Naturalclar, line 843: This lint rule does not exist for the oss version of recoil, which was causing lint error

PR #695, opened 2020-10-30
Update isArrayBufferView

EricPKerr:
This updates the isArrayBufferView method to use the types as a Set instead of linearly scanning the array.

### DIFF
--- a/src/util/Recoil_isArrayBufferView.js
+++ b/src/util/Recoil_isArrayBufferView.js
@@ -10,26 +10,8 @@
  */
 'use strict';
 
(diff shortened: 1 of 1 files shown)


drarmstr:
Hmm, actually, seems like we should just use `ArrayBuffer.isView(x)`...

EricPKerr:
Good call - updated to use that instead

drarmstr:
Updated with #702 which also removed some extra code we don't need anymore.  Thanks for caring about this @EricPKerr !

Issue #697, opened 2020-10-30
Type bug for Loadable in 0.1.1


wmain:
Reproduction: https://codesandbox.io/s/hidden-frost-o99qc

`Loadable`'s interface now expects a state of `error`, but actually returns a state of `hasError`.

PR #698, opened 2020-10-30
Fix typescript bug for the Loadable interface (#697)

wmain:
Summary: The error state of Loadable should be "hasError" not "error."


wmain:
Looks like this is now a dupe of https://github.com/facebookexperimental/Recoil/pull/699/files

PR #702, opened 2020-10-30
Simplify ArrayBufferView check

drarmstr:
Summary:
Simplify the `ArrayBufferView` check by using [ArrayBuffer.isView()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView) which should be supported on all browsers.

Try to avoid breaking React Native...

Differential Revision: D24655610


EricPKerr:
Polyfill if needed: https://github.com/jvilk/BrowserFS/issues/193#issuecomment-319771804

drarmstr:
> Polyfill if needed: [jvilk/BrowserFS#193 (comment)](https://github.com/jvilk/BrowserFS/issues/193#issuecomment-319771804)

Thanks for the link if anyone needs it.  Coverage looks pretty decent now: 
image

PR #705, opened 2020-10-31
Add type for persistence_UNSTABLE in AtomOptions

junhoyeo:


junhoyeo:
Oops I didn't notice this feature was changed to using `effects_UNSTABLE`.

PR #713, opened 2020-11-05
Add useGetRecoilValueInfo() hook

drarmstr:
Summary: Add `useGetRecoilValueInfo_UNSTABLE()` hook which acts similar to `getInfo_UNSTABLE()` method on `Snapshot`, except peeks the info for the current state.

Differential Revision: D24718803


drarmstr:
Document in #744

PR #714, opened 2020-11-05
report subscribing components

drarmstr:
Summary:
Report components which are subscribing to atoms and selectors via `useGetRecoilValueInfo()`.

Note that `Snapshot#getInfo()` will not report any subscribing components at this time.

This currently reports a string of the component name.  But, we likely want to change that to something more specific/useful.

Reviewed By: maxijb

Differential Revision: D24718812


drarmstr:
Document in #744

PR #717, opened 2020-11-06
fix: snapshot variable is not defined

wabi91:
**Summary**
There is an error because the variable is wrong.
`initialSnapshot`, `testSnapshot` and `mySnapshot` should be changed correctly.


drarmstr:
Thanks @wabi91 !

PR #749, opened 2020-11-18
Avoid re-rendering components if selector values don't change

drarmstr:
Summary:
An optimization that will avoid re-rendering React components if selector values they subscribe to don't change (based on reference equality).

Currently, changing an atom to the same value (based on reference equality) will not propagate any downstream updates.  However, any updates to a selector would propagate updates to all downstream selectors and re-render all subscribing components.

This is only a partial optimization because downstream selectors will still re-evaluate.  However, the React component itself will see that the final value has not changed and will suppress re-rendering.  This is also only a partial fix because it only limits suppression based on reference-equality, and does not allow for a user-defined equality check.

Additionally - In order to compare the new and old values before re-rendering, the subscribing component will evaluate the selector before deciding to re-render.  It gets the value again during rendering, however the value will be cached in `atomValues`, so this should be cheap.  By moving this evaluation earlier, it may actually help "pre-fetch" asynchronous requests before rendering starts.

Differential Revision: D25040920

### DIFF
--- a/src/hooks/Recoil_Hooks.js
+++ b/src/hooks/Recoil_Hooks.js
@@ -40,6 +40,7 @@ const differenceSets = require('../util/Recoil_differenceSets');
 const expectationViolation = require('../util/Recoil_expectationViolation');
 const filterMap = require('../util/Recoil_filterMap');
 const filterSet = require('../util/Recoil_filterSet');
+const gkx = require('../util/Recoil_gkx');
 const invariant = require('../util/Recoil_invariant');
 const mapMap = require('../util/Recoil_mapMap');
 const mergeMaps = require('../util/Recoil_mergeMaps');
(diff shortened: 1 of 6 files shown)


drarmstr:
Note that while this diff was merged it is currently behind a feature flag.

salvoravida:
maybe related https://github.com/facebookexperimental/Recoil/pull/825

src/hooks/Recoil_Hooks.js
```diff
@@ -285,42 +286,57 @@ function useRecoilValueLoadable_MUTABLESOURCE<T>(
...
     },
-    [recoilValue, storeRef, componentName],
+    [storeRef, recoilValue, componentName, getLoadable],
   );
 
-  return useMutableSource(useRecoilMutableSource(), getValue, subscribe);
+  const source = useRecoilMutableSource();
+  const loadable = useMutableSource(source, getLoadableWithTesting, subscribe);
+  const prevLoadableRef = useRef(loadable);
+  useEffect(() => {
```
escaton, line 336: @drarmstr why not `layoutEffect`? Postponed update of the ref could cause some propblem, don't you think?

pthieu:
@drarmstr did this ever get released?

chriszrc:
@pthieu looks like yes - https://recoiljs.org/docs/api-reference/core/RecoilEnv/

Screenshot 2024-05-03 at 1 03 55 PM

PR #763, opened 2020-11-26
Fix React Native build

mondaychen:
- revert 8e1e2ee
- move Reactbatchedupdates.js and Reactbatchedupdates.native.js to polyfill to avoid reaper


build/index.js
```diff
@@ -0,0 +1,6 @@
+const Recoil = require('../src/Recoil_index');
```
drarmstr, line 1: Needs copyright header

build/index.native.js
```diff
@@ -0,0 +1,6 @@
+const Recoil = require('../src/Recoil_index');
```
drarmstr, line 1: Needs copyright header

drarmstr: It's not clear from the GitHub PR where this file lives in `www` and if it may also be a candidate for the reaper?

setupJestMock.js
```diff
@@ -1,3 +1,5 @@
 const Promise = require('promise-polyfill');
```
drarmstr, line 1: Needs copyright header

src/core/Recoil_Batching.js
```diff
@@ -10,9 +10,14 @@
...
 const {batchStart} = require('../core/Recoil_RecoilValueInterface');
-const {unstable_batchedUpdates} = require('../util/Recoil_ReactBatchedUpdates');
 
-let batcher = unstable_batchedUpdates;
+let batcher = function (_func) {
+  throw new Error('Unexpected Fatal Error: undefined batcher function');
+};
+// for testing, use default batcher defined in setupJestMock.js
+if (global && global._jestDefaultBatcher) {
+  batcher = global._jestDefaultBatcher;
```
drarmstr, line 19: Could we avoid `setupJestMock.js` and inside the default batcher just check if running under Jest and use the `unstable_batchedUpdates()` if so, otherwise throw the error?  Or maybe just call `Recoil.setBatcher()` in some Jest setup function?

drarmstr (commented):
This does look cleaner than the previous approach!

build/index.native.js
mondaychen: don't worry -- this will be in scripts like other build scripts

mondaychen:
updated per discussion with @drarmstr

Issue #422, opened 2020-07-01
Force re-evaluation of selectors


jcrowson:
Hey Recoil team,

What’s the standard for calling an async Selector on, say, an onClick event?

Would it be to tie the Selector to an Atom, and then change the value of the Atom?

drarmstr:
@jcrowson - Could you provide more context?  If you're just asking about reading a selector in a callback, vs during the render, you can do that as follows.  Motivations to do this might be to avoid an expensive query until the callback, prefetching, etc.  See the [`useRecoilCallback() docs`](https://recoiljs.org/docs/api-reference/core/useRecoilCallback)

```jsx
function MyButton() {
  const onClick = useRecoilCallback(({snapshot}) => async () => {
    const results = await snapshot.getPromise(expensiveQuery);
    alert(results);
  });

  return (
    <div>
      <button onClick={onClick}>My Button</button>
    </div>
  );
}
```

Is this what you're asking about?

NethraRaghu:
I have similar doubt, 

I have a selector that does the API call and I would like to call it onClick of a button.

`const [var, setter] = useRecoilStateLoadable(<selector>);`
If I add this in onclick function it goes to an infinite loop.
If I add it in parent function it gets called before onclick. 

Whats the best practice and can you provide an example?

NethraRaghu:
Hi, Is it possible to get a loading state, until the promise is resolved?

jacques-blom:
@drarmstr, I think @jcrowson is asking how you can rerun an async selector imperatively? An example use case would be when a user clicks refresh button.

I suppose the general idea is that you only need to fetch your data when your selector's dependencies change, but it makes sense that you may want to refetch data from the server in case it has changed.

As far as I know, the only way to do this would be to, as you say, update an atom value.

Code example: https://codesandbox.io/s/recoil-refetch-data-bqjp6?file=/src/App.tsx

drarmstr:
Correct, selectors represent derived state.  The selector `get` function is logically "idempotent," that is should always provide the same value for a given set of dependency values or selectorFamily parameters.  It may be cached or re-executed.  Only use selectors for queries where the query results will not change for the lifetime of the app.  If you are trying to have local state as a cache of some remote mutable state, then atoms can be used instead.  Effects can subscribe atoms to changes in other state, see [these examples](https://recoiljs.org/docs/guides/asynchronous-state-sync)

jcrowson:
[...]

PR #764, opened 2020-11-26
docs: help prevent "selector without dependencies" testing issues

jeysal:
I ran into #422/#650, costing me a lot of time, when writing tests for a selector that
retrieves data from a mocked fetch function that returns different values in different test cases.
Being confused how there can be any state persisting across a re-mount of `RecoilRoot`,
this is one of the documentation locations where I would have found the issue quicker.

### DIFF
--- a/docs/docs/api-reference/core/RecoilRoot.md
+++ b/docs/docs/api-reference/core/RecoilRoot.md
@@ -5,6 +5,8 @@ sidebar_label: <RecoilRoot />
 
 Provides the context in which atoms have values. Must be an ancestor of any component that uses any Recoil hooks. Multiple roots may co-exist; atoms will have distinct values within each root. If they are nested, the innermost root will completely mask any outer roots.
 
+Note: This does **not** mean that `RecoilRoot`s are always completely isolated from each other. For example, a selector that does not depend on any atom may have its result cached across roots. This can cause problems if the selector reads external resources, especially in tests that only render a new `RecoilRoot` for every test, but do not [reset](/docs/guides/asynchronous-data-queries#use-a-request-id) the selector in any way.
+
 ---
 
 **Props**:


docs/docs/api-reference/core/RecoilRoot.md
```diff
@@ -5,6 +5,8 @@ sidebar_label: <RecoilRoot />
 
 Provides the context in which atoms have values. Must be an ancestor of any component that uses any Recoil hooks. Multiple roots may co-exist; atoms will have distinct values within each root. If they are nested, the innermost root will completely mask any outer roots.
 
+Note: This does **not** mean that `RecoilRoot`s are always completely isolated from each other. For example, a selector that does not depend on any atom may have its result cached across roots. This can cause problems if the selector reads external resources, especially in tests that only render a new `RecoilRoot` for every test, but do not [reset](/docs/guides/asynchronous-data-queries#use-a-request-id) the selector in any way.
```
drarmstr, line 8: Adding some clarification may be helpful, but some corrections here:  It does not matter if the selectors have dependencies on other atoms/selectors or not for this.  It is also not a *problem* if the user is properly following the API contract when implementing their selectors to be properly "idempotent" (returning the same value for the same set of inputs/dependencies with no side-effects except for caching/logging).  Also, the term "reset" means to reset an atom/selector to the default value, which is distinct from the "refresh" pattern described in the guide.

drarmstr (changes requested):

docs/docs/api-reference/core/RecoilRoot.md
jeysal: I don't view the first two points as corrections.

> It does not matter if the selectors have dependencies on other atoms/selectors or not for this.

The referenced sentence starts with "For example" - it does not claim that there are no other situations in which this occurs, such as when a selector depends on an atom but the atom happens to have the same value across roots.

> It is also not a _problem_

It is not stated that Recoil's caching is a problem, it is stated that Recoil's caching in conjunction with certain user code can *cause* problems. This does not make a statement about either the former or the latter being at fault.

I agree that the wording reset is unfortunate, given that Recoil already uses it to describe one of its concepts. Substituting it with "refresh" would make sense and leave the meaning of the sentence intact.
Let me know if you want me to make this or other minor changes, however if as it appears you disagree with the phrasing of the paragraph overall, it'd probably be better if you rephrase it yourself from a maintainer POV, since it's barely worth bikeshedding any more over this few words :sweat_smile:

drarmstr:
Wordsmithed in #802

Issue #762, opened 2020-11-25
Async selector family set function


Khachatour:
Hello dear Recoil team! Using Recoil for the production application and loving it so far. Just one specific clarification.

In turns out that I can't have a async `selectorFamily` set function, example below

```js
const filtersSelector = selectorFamily({
  key: 'filtersSelector',
  get: (filterField: string) => ({ get }) => get(filtersState)[filterField],
  set: (filterField: string) => async ({ set, get }, newVal) => {
    const filterTerms = {
      ...get(filtersState),
      [filterField]: newVal,
    } as FilterTypes;

    const setter = (prevValue: FilterTypes) =>
      ({ ...prevValue, [filterField]: newVal } as FilterTypes);

    set(filtersState, setter);

    const data = await getPhoneNumbers(filterTerms);

    set(phoneNumbersState, data);
  },
});
```

This code snippet works this way, the `filtersState` set function works and sets new filter terms, after that the `getPhoneNumbers` request works (I can see it in the network tab) but the last `phoneNumbersState` set function does not change the state even though if I write 

```js
  set(phoneNumbersState, updateFunc);
```

the `updateFunc` gets invoked but I can't see state updates after that

Thanks in advance!

drarmstr:
[Selectors](https://recoiljs.org/docs/api-reference/core/selector) and [selector families](https://recoiljs.org/docs/api-reference/utils/selectorFamily) do not currently allow async `set()` methods.  Flow or TypeScript *should* give an error that you are trying to return a `Promise` from a `void` function.. 

The current mechanism to support async sets is to just use the setter from [`useRecoilState()`](https://recoiljs.org/docs/api-reference/core/useRecoilState) or [`useSetRecoilState()`](https://recoiljs.org/docs/api-reference/core/useSetRecoilState) asynchronously. Or, for transactions with multiple atoms or abstractions you can use [`useRecoilCallback()`](https://recoiljs.org/docs/api-reference/core/useRecoilCallback) to asynchronously set atoms.  E.g.:

```jsx
function useFilters(filterField: string) {
  const [filters, setFilters] = useRecoilState(filtersState);
  const setPhoneNumbers = useSetRecoilState(phoneNumbersState);
  const setFilter = useCallback(newValue => {
    setFilters(prevFilters => {
      const filterTerms = {...prevFilters, [filterField]: newValue});
      setFilters(filterTerms);
      const data = await getPhoneNumbers(filterTerms);
[...]

PR #777, opened 2020-12-01
Add error reporting for attempted async selector set

drarmstr:
Summary: Recoil does not support async selector sets.  However, as it does support async gets many users expect async sets.  Flow reports a static error if an `async` function is used, TypeScript should as well.  Though users may not have or notice this error.  This diff throws an error if callbacks in the `set` are attempted to be used asynchronously.  However, those errors may not be caughts and be lost.  So, we also detect if an `async` function was used by checking if the user's set returns a `Promise` and throws an error for that as well.

Differential Revision: D25250686

### DIFF
--- a/src/recoil_values/Recoil_selector_NEW.js
+++ b/src/recoil_values/Recoil_selector_NEW.js
@@ -879,10 +879,17 @@ function selector<T>(
     function mySet(store, state, newValue): [DependencyMap, AtomValues] {
       initSelector(store);
 
+      let syncSelectorSetFinished = false;
       const dependencyMap: DependencyMap = new Map();
       const writes: AtomValues = new Map();
 
       function getRecoilValue<S>({key}: RecoilValue<S>): S {
+        if (syncSelectorSetFinished) {
+          throw new Error(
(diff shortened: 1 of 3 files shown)


jjalonso:
Will async sets on selector will never be implemented? 
its quite unfair that it is for gets only, some people like me need to derivate data using promise, one example is this.

Imagine I have two kind of events that im subscribed on externally, channelAdded, channelRemoved, on channelAdded and channelRemoved i get as argument a channel instance from an 3rd party lib, and need to call some methods that return promises, like getMembers() etc, in order to serialize, 

Then we endup having a bidirectional selector get/set where the get return alls channels and the set have a switch case on the argument to evaluate value.type 'ADDED' or 'REMOVED'  in order to do something with value.payload but just the added need to serialize. Ideally we should need to have the derivate data logic inside the setter and not doing out previously the set call, 

currently with the current implementation that doesnt support async set, we need to serialise out and send to the set the channel serialize in case of added and the channel without serialize in case of remove.

drarmstr:
@jjalonso - Continuing discussion in #762

PR #787, opened 2020-12-10
[docs] async fix typo userRecoilValue

AndersDJohnson:
Recoil is looking very cool! Excited to use it where traditionally we might have reached for Redux or custom context...
I was just going through the latest docs and noticed a small typo on async queries showing `userRecoilValue` instead of `useRecoilValue`.


mondaychen:
Great catch! Thanks!

Issue #41, opened 2020-05-16
Object as an atom will cause re-render even when use selector?


Albert-Gao:
```javascript
import * as React from "react";
import { atom, useRecoilValue, useSetRecoilState, selector } from "recoil";
import "./App.css";

const valueState = atom({
  key: "value",
  default: { a: 1, b: 2 },
});

const selectA = selector({
  key: "selectA",
  get: ({ get }) => get(valueState).a,
});

function ValueChild() {
  const { a } = useRecoilValue(valueState);

  return <div>useRecoilValue a:{a}</div>;
}

function SelectorChild() {
  const a = useRecoilValue(selectA);

  return <div>useRecoilValue + Selector a:{a}</div>;
}

function App() {
  const setValues = useSetRecoilState(valueState);
  const onlyChangeB = () => {
    setValues((v) => ({ ...v, b: 3 }));
  };

  return (
    <div className="App">
      <ValueChild />
      <SelectorChild />
      <button onClick={onlyChangeB}>update</button>
    </div>
  );
}

export default App;
```

Bootstrapped a new create-react-app and play with it.

I declared an atom as an object `{ a: 1, b: 2 }`

with 2 components which consume it:

- `<ValueChild>` only subscribe to property `a`
- `<SelectorChild>` only subscribe to property `a` with selector

**Expect:**
change property `b` should not trigger re-render of components who subscribe to property `a`

**Actual behavior**
when I click the button to update only `b`'s value, both components will get re-rendered.

Anything I am missing here? Thanks. Very nice API and idea. I love the work. Tried to replace redux/mobx with this. But this will be a blocker

drarmstr:
Subscriptions are based on individual atoms and selectors. Currently, all downstream subscribers of an atom or selector will update when a dependency updates. If you wish to have your component only update when a or b is updated, then they can be modeled as separate atoms.

Internally, we're working on a mechanism to limit updates when the values have not changed.  But, we're planning to really clean up that API before publishing it.

Albert-Gao:
Thanks, so currently, atom means primitive value? or array(consider it is in the todo example)?

drarmstr:
Atoms or Selectors can represent either primitives, arrays, or complex objects.  But, subscriptions are only handled at an atom/selector granularity.  We've found it a best-practice to keep atom values rather simple.

rwieruch:
Thanks for this library! Really love to play around with it.

I am currently working on this example to demonstrate Recoil:

```
[...]

PR #808, opened 2020-12-25
Document why certain selectors cause an update when there are no apparent changes

frankiesardo:
I've been puzzling why the Tutorial example was re-rendering for no apparent reason and I finally found this explanation #41 

I wrongly assumed that selectors would behave in a react-ish way and do not cause a re-render if their value does not change. Others might fall in the same trap, I think adding a line of explanation would be useful.

Superb work on the framework, btw, and I'm really looking forward to be using the API mentioned here https://github.com/facebookexperimental/Recoil/issues/41#issuecomment-629601674. We would be happy even if it's just released as an `_UNSTABLE` preview.

### DIFF
--- a/docs/docs/basic-tutorial/selectors.md
+++ b/docs/docs/basic-tutorial/selectors.md
@@ -142,6 +142,8 @@ function TodoListStats() {
 }
 ```
 
+Note that in the current implementation `TodoListStats` will re-render even when there are no apparent changes to the `todoListStatsState`, for example when you edit a `TodoItem`. This is because all downstream subscribers of an atom or selector will update when a dependency updates.
+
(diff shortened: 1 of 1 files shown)


drarmstr:
Atoms will not cause a re-render if set to the same value.  This will also no longer be true for selectors with optimizations like #749, though note that is currently behind a feature flag.

frankiesardo:
Super, thanks! I know Recoil's design encourages to have many little atoms, each one with their own state, but some of us prefer to keep it all in one place and consider selectors as "materialised views" of the main atom. Glad this is now possible.

salvoravida:
maybe related to https://github.com/facebookexperimental/Recoil/pull/825 ?
@frankiesardo

Issue #565, opened 2020-09-09
unnecessary re-renderings


suren-atoyan:
Please examine [this](https://codesandbox.io/s/not-so-clear-re-renderings-ltz3n) code snippet.

The content of the snippet is simple and looks like this:

```javascript
import React from "react";

import { atom, useRecoilValue } from "recoil";

const exampleState = atom({
  key: "exampleState",
  default: "some value"
});

export default function App() {
  const value = useRecoilValue(exampleState);

  console.log("check render");  // it logs 4 times

  return <div>{value}</div>;
}
```

There is nothing special (I guess), one `atom` and one `useRecoilValue` inside the only component. The problem is that the `console.log` that we have in the component logs 4 times 

`useRecoilValue` (without updating the atom) re-renders the component (which uses it) 4 times in the beginning. If we change our code a little bit to have further updates of the atom we can see 2 more logs after each update 😕 

```javascript
import React, { useEffect } from "react";

import { atom, useRecoilState } from "recoil";

const exampleState = atom({
  key: "exampleState",
  default: "some value"
});

export default function App() {
  const [value, setValue] = useRecoilState(exampleState);

  useEffect(() => {
    const timerId = setInterval(() => {
      setValue("other value");
    }, 2000);

    return () => clearInterval(timerId);
  }, [setValue]);

  console.log("check render"); // it logs 4 times at the beginig
  // plus 2 more logs after atom's every update

  return <div>{value}</div>;
}
```

[Here](https://codesandbox.io/s/not-so-clear-re-renderings-2-gb7zl?file=/src/App.js:0-574) is the snippet for this one.

In the ideal (normal) case I suspect to see one log (with 'some value') and after about 2 seconds one more log (with 'other value') and that's all because further updates don't change the value.

I've found [this](https://github.com/facebookexperimental/Recoil/issues/314) issue and some others related to it, but they are about `selector` optimization. Here we have not any `selector`, there is just one `atom` and its usage in the component (via `useRecoilValue`/`useRecoilState`)

drarmstr:
Duplicate of #307

@suren-atoyan - React strict mode will perform extra renders to help catch side-effects (#75) are you using that?  We also have added some improvements to reduce extra renders with Recoil.  (#307) Do you see similar behavior with the Recoil master branch and without React strict mode?

suren-atoyan:
[...]

Issue #307, opened 2020-06-11
render twice on component mount (useRecoilValue, useRecoilState) no StrictMode


NikitaBazhenov:
Hello,

I am really excited about this project, I've played around and just notice that my component was rendered twice after it mounted even without StrictMode(in production) #75 

I've created a small example without StrictMode enabled on code sandbox: 
https://codesandbox.io/s/recoil-itit-double-render-jzbun

just wondering what I am doing wrong
I appreciate any help!

Shmew:
I've also noticed it always renders twice on initialization of a component that subscribes to a recoil value.

elis-k:
Just noticed that it rendered 4 times on initalization of a component for me.

NikitaBazhenov:
@elis-k probably you are run your app in development env and you are using <React.StrictMode> on root level in your app, please check this issue related to StrictMode: [facebook/react#15074](https://github.com/facebook/react/issues/15074)
so if you are in StrictMode you will see 4 times rendering in dev (2 times React.StrictMode + 2 times useRecoilValue) and in production build, StrictMode is doesn't work, so it will be 2 times because of useRecoilValue) 😞

Shmew:
Yeah I've tested renders and it's only on the initialization, after that it's a single render per committed update that's subscribed by the component.

semvis123:
Does someone have a fix for this? it breaks my code.

drarmstr:
@semvis123 - How is this breaking your code?

semvis123:
@drarmstr It causes my function to run twice, which is not good for performance, and if API calls are placed inside those get called twice.

drarmstr:
React render functions shouldn’t have side effects, so this should only be an optimization not a correctness issue. That’s why the React strict mode does extra renders to help test for side effects.

semvis123:
Where should I put my function, so it only runs once?

drarmstr:
Check out ‘useEffect()’ for side effects.

semvis123:
Thanks, I will definitely try that.

drarmstr:
Recent optimizations should remove the extra render from Recoil during mount.  This should be available in master; we're working on fixing an issue and then it will be release with 0.0.11.

smith-chris:
Hey, I believe this is still happening with lates `0.1.1` and `0.0.11` mentioned above. `React.StrictMode` disabled ofc. Any ideas? 

[...]

PR #820, opened 2021-01-05
Fix initial double render

salvoravida:
Fix initial double render

issue related:

https://github.com/facebookexperimental/Recoil/issues/565

https://github.com/facebookexperimental/Recoil/issues/307

Issue CodeSandBox
https://codesandbox.io/s/recoil-legacy-double-render-30bjc

Fix CodeSandBox
https://codesandbox.io/s/recoil-legacy-double-render-fixed-2oxfj


drarmstr:
Great idea to extend the #749 optimization to also resolve the initial double-render issue!  I think we'd want to restrict this behind the same GK feature flag until we are confident on the safety of the approach.  We can try to stress test this approach under that.

Issue #767, opened 2020-11-28
effects_UNSTABLE not triggering when using RecoilRoot initializeState 


abacaj:
RecoilRoot using initializeState:
```
      <RecoilRoot
        initializeState={({ set }) => set(domainState, props.savedDomainState)}
      >
        <AppRoot root={props.root} />
      </RecoilRoot>
```

The Atom using an effect:
```
const domainState = atom({
  key: "domainState",
  default: null,
  effects_UNSTABLE: [syncDomainStorage],
});
```

The function: syncDomainStorage is never triggered if I use `initializeState`, if I remove initializeState the function is called when the atom is updated.

drarmstr:
@abacaj - I added a unit test for the interaction of atom effects and `<RecoilRoot>`'s `initializeState` prop in #771 and it appears to be working.  Note that with the current implementation the effect is run first, then the global `initializeState` runs and thus takes precedence setting the initial atom value.  Perhaps that is the discrepancy you are observing?  Do you have a reproducer for your issue on `codesandbox.io`?

abacaj:
@drarmstr thanks for the quick reply.

Please see example here:
https://codesandbox.io/s/confident-framework-ib1kf?file=/src/index.js

I am appending to the DOM every time the effect is supposed to trigger:
```
function handleStateSideEffects({ onSet, setSelf, trigger }) {
  const textnode = document.createTextNode(`${trigger} - effect called`);
  const node = document.createElement("div");
  node.appendChild(textnode);
  document.body.appendChild(node);
}

export const appState = atom({
  key: "appState",
  default: null,
  effects_UNSTABLE: [handleStateSideEffects]
});
```

There is only two times that it is appended. Any time you click the button, it should be appending to the DOM - unless effects are supposed to perform differently?

abacaj:
@drarmstr thanks for the quick reply.

Please see example here:
https://codesandbox.io/s/confident-framework-ib1kf?file=/src/index.js

I am appending to the DOM every time the effect is supposed to trigger:
```
import { atom } from "recoil";

function handleStateSideEffects({ onSet, setSelf, trigger }) {
  onSet((newState) => {
    const textnode = document.createTextNode(`${newState} - effect called`);
    const node = document.createElement("div");
    node.appendChild(textnode);
    document.body.appendChild(node);
  });
}

export const appState = atom({
  key: "appState",
  default: null,
  effects_UNSTABLE: [handleStateSideEffects]
});
```

[...]

PR #828, opened 2021-01-12
More examples of Local Storage Persistence

mthines:
Asynchronous / Promise examples of Local Storage Persistence

https://github.com/facebookexperimental/Recoil/issues/767

### DIFF
--- a/docs/docs/guides/atom-effects.md
+++ b/docs/docs/guides/atom-effects.md
@@ -212,9 +212,9 @@ const syncStorageEffect = userID => ({setSelf, onSet, trigger}) => {
 
 ## Local Storage Persistence
 
-Atom effects can be used to persist atom state with [browser local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).  Note that the following examples are simplified for illustrative purposes and do not cover all cases.
+Atom effects can be used to persist atom state with [browser local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage). Since `localStorage` is synchronous, we can retrieve the items without `async` `await` or `Promise`.
 
-### Synchronous
+Note that the following examples are simplified for illustrative purposes and do not cover all cases.
 
 ```jsx
 const localStorageEffect = key => ({setSelf, onSet}) => {
@@ -241,14 +241,17 @@ const currentUserIDState = atom({
 });
 ```
 
-### Asynchronous
+## Asynchronous Storage Persistence
 
-If your persisted data needs to be `await`ed, you can either create an `asynchronous` function and call that or use a `promise` in the setSelf function 
+If your persisted data needs to be retrieved asynchronously, you can either create an `asynchronous` function and call that or use a `Promise` in the `setSelf()` function.
 
-- Using the promise approach you'll be able to wrap the components inside of the `<RecoilRoot/>` with a `<Suspense/>` component to show a fallback state while waiting for `Recoil` to load the persisted values.
-- This is not possible with the async await approach.
+### `Promise`
 
-#### Promise
+By synchronously calling `setSelf()` with a `Promise` you'll be able to wrap the components inside of the `<RecoilRoot/>` with a `<Suspense/>` component to show a fallback state while waiting for `Recoil` to load the persisted values.
+
+The `<Suspense/>` fallback will be shown until the `setSelf()` function has been called.
+
+Note that if the `atoms` later are "reset", they will revert to their default value, and not the initialized value.
 
 ```jsx
 const localStorageEffect = key => ({setSelf, onSet}) => {
@@ -279,29 +282,36 @@ const currentUserIDState = atom({
 ```
 
 
-#### Async Await
+### `Asynchronous` 
+
+Using an Asynchronous approach your `atoms`, `selectors` etc, will use their default value until the deferred `setSelf()` has been called to set the value.
+
+Unlike the `Promise` approach, it is therefore not possible to use `<Suspense/>` to show a fallback.
+
+This approach isn't just limited to `await`, but for any `async` usage of `setSelf()`, such as `setTimeout()`.
+
+### `AsyncLocalStorage`/`localForage` example
 
 ```jsx
 
-/** If there's a persisted value - set it on load  */
-const loadPersisted = async ({ key, setSelf }) => {
-  const savedValue = await localStorage.getItem(key);
+const localForageEffect = key => ({setSelf, onSet}) => {
+  /** If there's a persisted value - set it on load  */
+  const loadPersisted = async () => {
+    const savedValue = await localForage.getItem(key);
 
-  if (savedValue != null) {
-    setSelf(JSON.parse(savedValue));
-  }
-};
+    if (savedValue != null) {
+      setSelf(JSON.parse(savedValue));
+    }
+  };
 
-
-const localStorageEffect = key => ({setSelf, onSet}) => {
   // Load the persisted data
-  loadPersisted({ key, setSelf })
+  loadPersisted()
 
   onSet(newValue => {
     if (newValue instanceof DefaultValue) {
-      localStorage.removeItem(key);
+      localForage.removeItem(key);
     } else {
-      localStorage.setItem(key, JSON.stringify(newValue));
+      localForage.setItem(key, JSON.stringify(newValue));
     }
   });
 };
@@ -310,7 +320,7 @@ const currentUserIDState = atom({
   key: 'CurrentUserID',
   default: 1,
   effects_UNSTABLE: [
-    localStorageEffect('current_user'),
+    localForageEffect('current_user'),
   ]
 });
 ```
--- a/docs/docs/guides/atom-effects.md
+++ b/docs/docs/guides/atom-effects.md
@@ -214,12 +214,88 @@ const syncStorageEffect = userID => ({setSelf, onSet, trigger}) => {
 
 Atom effects can be used to persist atom state with [browser local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).  Note that the following examples are simplified for illustrative purposes and do not cover all cases.
 
+### Synchronous
+
 ```jsx
 const localStorageEffect = key => ({setSelf, onSet}) => {
   const savedValue = localStorage.getItem(key)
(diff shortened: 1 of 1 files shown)


mthines:
I've now signed the agreement. 🙂

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
 });
 ```
 
+### Asynchronous
+
+If your persisted data needs to be `await`ed, you can either create an `asynchronous` function and call that or use a `promise` in the setSelf function
```
drarmstr, line 246: `promise` -> `Promise`
setSelf -> `setSelf()`

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
 });
 ```
 
+### Asynchronous
```
drarmstr, line 244: `localStorage` is only sync afaik.  For async examples let's move out of the "Local Storage" section into another subsection and not reference `localStorage`.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
...
+});
+```
+
+
+#### Async Await
+
+```jsx
+
+/** If there's a persisted value - set it on load  */
+const loadPersisted = async ({ key, setSelf }) => {
```
drarmstr, line 287: Passing through the `setSelf()` callback might be confusing.  To help with clarity for documentation purposes maybe define this async function in the effect function so it can just use `key` and `setSelf()` directly.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
...
+  const savedValue = await localStorage.getItem(key);
```
drarmstr, line 288: See above comment re `localStorage` being sync

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
...
+  default: 1,
+  effects_UNSTABLE: [
+    localStorageEffect('current_user'),
+  ]
+});
+```
+
+
+#### Async Await
+
```
drarmstr, line 283: Clarify that with this approach, in contrast to the above, that it will use the default value until the async value was set.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
...
+  key: 'CurrentUserID',
+  default: 1,
+  effects_UNSTABLE: [
+    localStorageEffect('current_user'),
+  ]
+});
+```
+
+
+#### Async Await
```
drarmstr, line 282: This approach isn't limited to just `await`, but for any async usage of `setSelf()`, such as with `setTimeout()`.

drarmstr (changes requested):
Thanks @mthines for helping to improve our documentation!!

docs/docs/guides/atom-effects.md
mthines: That's very true. My original code was using `localForage` which is `async`, so it made sense in that case, but I converted that part to `localForage` since the other framework was related to my project and not recoil.

I'm uncertain how to write the documentation for this part then. Any suggestions?

docs/docs/guides/atom-effects.md
mthines: With documentation in mind, that makes sense. I'll do that.

docs/docs/guides/atom-effects.md
mthines: I clarified the pros/cons up at the asynchronous title, so you can read the difference together to make it easier to compare.

https://github.com/facebookexperimental/Recoil/pull/828/files#diff-33b7a54bcbd3c5dce0fc9f70d50de2e2a9260d9865c0a890252b2d92bb9d668eR248-R249

I can move each part down under each title as well, but I figure maybe it was unnecessary repetition.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
 });
 ```
 
+### Asynchronous
+
+If your persisted data needs to be `await`ed, you can either create an `asynchronous` function and call that or use a `promise` in the setSelf function 
+
+- Using the promise approach you'll be able to wrap the components inside of the `<RecoilRoot/>` with a `<Suspense/>` component to show a fallback state while waiting for `Recoil` to load the persisted values.
```
drarmstr, line 248: Let's not call this the "promise approach", as `Promise`s may be used in other approaches as well.  Maybe "`Synchronously call setSelf() with a Promise`".  Wordy, perhaps, but more explicit.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,80 @@ const currentUserIDState = atom({
...
+- This is not possible with the async await approach.
+
+#### Promise
+
```
drarmstr, line 252: Explicitly mention that this will use suspense to show a fallback while the state is pending the initial value.  If the atom is later "reset" then it will revert to the default value, not the initialized value.

docs/docs/guides/atom-effects.md
drarmstr: Maybe the "`asynchronously call setSelf()`" approach to again be more explicit.

Mention that the default atom value will be used until the deferred `setSelf()` is called to set the value.

docs/docs/guides/atom-effects.md
drarmstr: After reading through again I think it would be clearer to explain each approach's quirks in the sections themselves, thanks.

mthines: @drarmstr That's fine by me.
I'll move each part down :)

docs/docs/guides/atom-effects.md
mthines: Haha, I prefer it being more readable and explicit.

mthines:
Hey @drarmstr!

Sorry for taking so long getting back to this. I've had a crazy busy week 🏃 

I've updated the `atom-effects.md` based on the changes and suggestions you had.
Let me know if I've missed anything, or somethings still unclear.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,89 @@ const currentUserIDState = atom({
...
+By synchronously calling `setSelf()` with a `Promise` you'll be able to wrap the components inside of the `<RecoilRoot/>` with a `<Suspense/>` component to show a fallback while waiting for `Recoil` to load the persisted values.
+
+The `<Suspense/>` fallback will be shown until the `setSelf()` function has been called.
+
+Note that if the `atoms` later are "reset", they will revert to their default value, and not the initialized value.
+
+```jsx
+const localStorageEffect = key => ({setSelf, onSet}) => {
+  setSelf(
+    localStorage.getItem(key).then((savedValue) => {
```
drarmstr, line 259: `localStorage` only stores UTF-16.  Best to avoid `localStorage` completely for the async examples.  You could also just use `localForage` for this:

```
setSelf(localForage.getItem(key).then(savedValue =>
  savedValue != null
    ? JSON.parse(savedValue)
    : new DefaultValue() // Abort initialization if no value was stored
));
```

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,89 @@ const currentUserIDState = atom({
...
 
+## Asynchronous Storage Persistence
+
+If your persisted data needs to be retrieved asynchronously, you can either use a [`Promise`](#promise) in the `setSelf()` function or create an [asynchronous](#asynchronous) function and call that.
+
+### `Promise`
+
+By synchronously calling `setSelf()` with a `Promise` you'll be able to wrap the components inside of the `<RecoilRoot/>` with a `<Suspense/>` component to show a fallback while waiting for `Recoil` to load the persisted values.
+
+The `<Suspense/>` fallback will be shown until the `setSelf()` function has been called.
```
drarmstr, line 252: "`<Suspense>` will show a fallback until the `Promise` provided to `setSelf()` resolves."

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,89 @@ const currentUserIDState = atom({
...
+  const loadPersisted = async () => {
+    const savedValue = await localForage.getItem(key);
+
+    if (savedValue != null) {
+      setSelf(JSON.parse(savedValue));
+    }
+  };
+
+  // Load the persisted data
+  loadPersisted()
```
drarmstr, line 307: Missing semi-colon.

docs/docs/guides/atom-effects.md
```diff
@@ -239,6 +241,89 @@ const currentUserIDState = atom({
...
+  ]
+});
+```
+
+
+### Asynchronous
+
+Using an asynchronous function your `atoms`, `selectors` etc, will use their default value until the deferred `setSelf()` has been called to set the value.
+
+Unlike using a `Promise`, it is not possible to use `<Suspense/>` to show a fallback.
```
drarmstr, line 289: "Unlike initializing to a `Promise`, the atom's default value will be used initially, so `<Suspense>` will not show a fallback unless the atom's default is a `Promise` or async selector."

drarmstr (changes requested):
Thank you so much for iterating on this!  I think this should be the last round.

docs/docs/guides/atom-effects.md
mthines: Sorry, that's my bad.
We already talked about this, so simply a typo.
Thanks!

mthines: Resolved.

docs/docs/guides/atom-effects.md
mthines: Resolved.

docs/docs/guides/atom-effects.md
mthines: Resolved.

docs/docs/guides/atom-effects.md
mthines: Resolved.

PR #833, opened 2021-01-15
Updated typescript index.d.ts with typed key extending NodeKey

mthines:
I don't have the complete overview, and the bottom part doesn't look correctly typed, so maybe this is completely off, but like this I have typed keys if I pass it the function.


mthines:
74bd89d

PR #834, opened 2021-01-15
Updated typescript index.d.ts with typed key extending NodeKey

mthines:
I don't have the complete overview, and the bottom part doesn't look correctly typed, so maybe this is completely off, but like this I have typed keys if I pass it the function.

This will also result in a breaking change of the generic P type being after the newly implemented K type


mthines:
@drarmstr Is this something which make sense for you guys?
If not, I'll close the PR :)

mthines:
Closing as I didn't get a response, and then it's probably a personal preference rather than something that should be part of the project ☺️

PR #835, opened 2021-01-15
Remove all  unnecessary semi-colon from Recoil_Hooks.js

SoorajSahu:
There is no need for semi-colon in js after ES6


davidmccabe:
Thanks for your PR, but we use Prettier for code formatting.

PR #862, opened 2021-01-30
[docs] Update sidebars.js

escaton:
Some pages reference topics from next pages, reorder it to make things less confusing.


docs/sidebars.js
```diff
@@ -62,9 +62,9 @@ module.exports = {
           'api-reference/utils/constSelector',
           'api-reference/utils/errorSelector',
           'api-reference/utils/waitForAll',
-          'api-reference/utils/waitForAny',
-          'api-reference/utils/waitForNone',
           'api-reference/utils/noWait',
```
drarmstr, line 65: The `waitFor*()` entries are more related and should be kept together.  But, should be fine to move `noWait()` above those three.

drarmstr (changes requested):

PR #894, opened 2021-02-19
Fix loading failure on IE11

### LINKED ISSUE #893: Failed to load on IE11 with SCRIPT1047
When Recoil loaded on IE11, it failed to load js with the following error:

> SCRIPT1047: In strict mode, function declarations cannot be nested inside a statement or block. They may only appear at the top level or directly inside a function body.

and it indicates https://github.com/facebookexperimental/Recoil/blob/master/src/recoil_values/Recoil_atom.js#L226

skyisle:
This fixes #893


drarmstr:
@skyisle - Thanks for the fix!

PR #895, opened 2021-02-22
refactor(eslintrc): remove some rules from packages/recoil-devtools/.eslintrc

mallchel:
prepare the packages directory for eslint global rules


mallchel:
I tried to save some rules in packages/recoil-devtools/.eslintrc with the error value

PR #896, opened 2021-02-23
Allow toJSON types to be used with atomFamily and selectorFamily

### LINKED ISSUE #760: [Typescript] Suggestion: Allow types with a `toJSON` function to be used as serializable param for families.
Currently I can't use `Date` as parameter for family.
According to https://github.com/facebookexperimental/Recoil/blob/c6d786bb89c20730fec3068346c0b4f7265ea808/typescript/index.d.ts#L270-L273 the parameter needs to be primitive or an array of primitives.

As far as I can tell, this is because it needs to be serializable by `Recoil_stableStringify`:
https://github.com/facebookexperimental/Recoil/blob/ca888d3c3a55a3128279d0124ebca9002377770b/src/recoil_values/Recoil_selectorFamily.js#L31-L40

But `Recoil_stableStringify` handles objects with a `toJSON()` function:
https://github.com/facebookexperimental/Recoil/blob/ca888d3c3a55a3128279d0124ebca9002377770b/src/util/Recoil_stableStringify.js#L64-L71

So my suggestion is to add an interface supporting that as type:
```typescript
export type SerializableParam = 
	| Primitive 
	| { toJSON: () => string } // Add support for json serializable objects, such as Date
	| ReadonlyArray<SerializableParam> 
	| Readonly<{[key: string]: SerializableParam}>
	;
```

BenjaBobs:
This allows such use cases as using Date as params for a family, e.g. if you want to create a time slice of a range of data.
This fixes #760

### DIFF
--- a/src/recoil_values/__tests__/Recoil_atomFamily-test.js
+++ b/src/recoil_values/__tests__/Recoil_atomFamily-test.js
@@ -148,6 +148,18 @@ testRecoil('Works with parameterized default', () => {
   expect(get(paramDefaultAtom({num: 2}))).toBe(2);
 });
 
+testRecoil('Works with date as parameter', () => {
+  const dateAtomFamily = atomFamily({
+    key: 'dateFamily',
+    default: date => 0,
+  });
+  expect(get(dateAtomFamily(new Date(2021, 2, 25)))).toBe(0);
+  expect(get(dateAtomFamily(new Date(2021, 2, 26)))).toBe(0);
+  set(dateAtomFamily(new Date(2021, 2, 25)), 1);
+  expect(get(dateAtomFamily(new Date(2021, 2, 25)))).toBe(1);
+  expect(get(dateAtomFamily(new Date(2021, 2, 26)))).toBe(0);
+});
+
 testRecoil('Works with parameterized fallback', () => {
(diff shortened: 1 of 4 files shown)


drarmstr (commented):
I'm used to Flow now, assuming TypeScript uses inexact types by default?  I can add similar typing (though inexact object) to the Flow types.  Thank you for adding the static TypeScript test.  Would you be up for adding actual Jest unit tests in `Recoil_selectorFamily-test.js` and `Recoil_atomFamliy-test.js` for using `Date`?

BenjaBobs:
Yes, it just means anything that has a `toJSON` property that is a `function` that returns a `string` is valid.

Sure, I'll look into it. 👍

BenjaBobs:
@drarmstr I added the tests you suggested, and rebased onto master. As for the code style of the tests, I tried as best I could to follow the precedent set by you guys, but will be happy to fix any issues/discrepancies you may find.

drarmstr:
> @drarmstr I added the tests you suggested, and rebased onto master. As for the code style of the tests, I tried as best I could to follow the precedent set by you guys, but will be happy to fix any issues/discrepancies you may find.

Wonderful!  Importing now for landing.

PR #919, opened 2021-03-10
fix confused initial state in Batcher

lllllllqw:
Hello, I found that the state in the Batcher component has a little style problem. It might be better to change it to this

- change initial state in Batcher


drarmstr:
@lllllllqw - What was the style problem?

PR #952, opened 2021-04-02
Fix suppression of re-renders with Suspense

drarmstr:
Summary:
Fix the ability to suppress re-rendering React Components that depends on Recoil Selectors which evaluate to equivalent values via reference equality.

The problem was that when components observed a selector in a pending state it would suspend using Suspense.  This would cancel the execution of `useEffect()`'s and skip recording the previous selector values.  Then, if the selector reverted to the same value it had before the suspense, it wouldn't appear the value changed and the component could get stuck in Suspense.

Differential Revision: D27547654


drarmstr:
PR didn't auto-close with commit landing.

Issue #961, opened 2021-04-11
Loadable TypeScript inference issue


RayLuxembourg:
I am not sure if this a typescript limitation or not but it is very wrong the hasValue state displays both error and success O_O
According to the flow in the docs [loadable docs](https://recoiljs.org/docs/api-reference/core/Loadable)
hasValue should be inferred only as the expected object on success

```
        const { state, contents } = useRecoilValueLoadable(userSelector);
        case 'hasValue':
        currentRole = RoleUX[contents.role] || contents.role;

```

![](https://i.imgur.com/BGXLXSB.png)

After some investigation, I realize this happens when you destruct from useRecoilValueLoadable.

Regarding this interface 
```
interface BaseLoadable<T> {
  getValue: () => T;
  toPromise: () => Promise<T>;
  valueMaybe: () => T | void;
  valueOrThrow: () => T;
  errorMaybe: () => Error | void;
  errorOrThrow: () => Error;
  promiseMaybe: () => Promise<T> | void;
  promiseOrThrow: () => Promise<T>;
  map: <S>(map: (from: T) => Promise<S> | S) => Loadable<S>;
}
```
Shouldn't those conditional voids be declared on proper state type?
for example,
```
interface ValueLoadable<T> extends BaseLoadable<T> {
  state: 'hasValue';
  contents: T;
  valueMaybe: () => T;
}

```
and removed from BaseLoadable and that way the type can become more accurate?

BenjaBobs:
To add to that, having `valueMaybe` return type `undefined` instead of `void` would make it easier to use with the `?.` operator.

drarmstr:
Sounds like some nice improvements, thanks!

PR #966, opened 2021-04-14
Loadable typing improvements

drarmstr:
Summary:
Some improvements for the typing of `Loadable`'s that address #961 and other issues:

# Flow
* When the state of the `Loadable` is refined refine the "maybe" accessors in addition to the type of the `contents`.

# TypeScript
* When the state of the `Loadable` is refined refine the "maybe" accessors in addition to the type of the `contents`.
* The error state of a `Loadable` should be `any` instead of `Error` as any type may be thrown
* Add missing `is()` method to check for `Loadable` equivalence.
* Return `undefined` instead of `void` to allow for `loadable.valueMaybe()?.xxx()`

Plus, a minor optimization for most `Loadable` methods to avoid checking the value of the `state`.

Differential Revision: D27779853

### DIFF
--- a/src/adt/Recoil_Loadable.js
+++ b/src/adt/Recoil_Loadable.js
@@ -35,7 +35,7 @@ export type LoadablePromise<+T> = Promise<
   ResolvedLoadablePromiseInfo<T> | Canceled,
 >;
 
-type Accessors<T> = $ReadOnly<{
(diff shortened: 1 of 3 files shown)


drarmstr:
These changes landed, not sure why the PR wasn't automatically closed..

PR #989, opened 2021-04-24
Add getCallback() to selector interface

drarmstr:
Summary:
Occasionally, one would like a selector to return an object that contains callbacks which can access Recoil state.  Add a `getCallback()` callback to the selector evaluation interface in addition to `get()` in order to support this.

When evaluating a selector `get()` will synchronously get the value of the dependency and subscribe the selector to update if that dependency changes.  `getCallback()`, will return a callback which has access to a snapshot of Recoil state at the time the snapshot was executed.  This means there is no subscription to re-evaluate the selector.  This matches the convention established by `useRecoilCallback()`, only within a selector.  Similarly, callback cannot be called while the selector is being evaluated.  It can only be used later asynchronously to get a `Promise` or `Loadable` of the value at the time of the callback.

An example when this is useful is if a user would like to use a selector to generate object which represent menu items.  The menu items may then have callbacks which access Recoil state.

```
const menuItemState = selectorFamily(
  'MenuItem',
  (itemID: number) => ({get, getCallback}) => {
    const name = get(itemNameQuery(itemID));
    const onClick = getCallback(({snapshot}) => async () => {
      const info = await snapshot.getPromise(itemInfoQuery(itemID));
      displayInfoModal(info);
    });
    return {
      title: `Show Info for ${name}`,
      onClick,
    };
  },
});
```

It is left as an exercise to the future if we would like to allow these callbacks to also set Recoil state.

Please refer to the original [RFC](https://fb.quip.com/si45AN3C6vpm) from habond

Differential Revision: D27985112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
     - On initial render when not using React Concurrent Mode (#820)
 - Improved TypeScript and Flow typing for `Loadable`s (#966)
 - Added override prop to RecoilRoot
+- Add getCallback() to selector evaluation interface
 - Fix not calling onSet() handler triggered from a setSelf() in onSet() for Atom Effects (#974, #979)
 
 ## 0.2.0 (2021-3-18)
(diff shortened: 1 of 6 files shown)


KestutisMa:
recoil version 3.1, i got error "Property 'getCallback' does not exist on type '{ get: GetRecoilValue; }'.ts(2339)"

drarmstr:
@KestutisMa - Thanks, looks like it was properly added to the TypeScript typing for `selector()`, but not `selectorFamily()`.  Will get a fix in.  It should work, just have the TypeScript error.

Issue #774, opened 2020-12-01
Setting atom from an effect concurrently with another update loses the value


idubrov:
Potentially, similar to #728.

Setting atom value inside an effect (via `setSelf`) while another update is happening loses the state change. Delaying `setSelf` via `window.requestAnimationFrame` solves the issue.

Here is the reproduction case I have: https://codesandbox.io/s/xenodochial-firefly-wny9s?file=/src/App.js

1. Click "Navigate!". Observe that "State A" value is empty and "State B" is "hello".
2. Now reload page in the embedded browser with `?delay` added to the URL. 
3. Click "Navigate!". Observe that "State A" value is now correct ("from effect") and "State B" is "hello".

drarmstr:
Please note that calling `setSelf()` from `window.requestAnimationFrame()` will first initialize the atom to the default value, then asynchronously update the value.  Calling `setSelf()` synchronously in the effect will initialize that atom to that value for the first render, which is important for things like SSR and rendering the initial state without potential flicker.

idubrov:
Sure. However, if I don't set stateB, it works as expected. Which makes it very confusing behavior. It only shows up if both are true:

1. We have data ready in our effect (so we call setSelf immediately).
2. At the same time, we set some other, unrelated state as well.

drarmstr:
Found the issue, thanks for the great repro @idubrov!  I have a fix in the works, though may be delayed through reviews due to holidays..

mdlavin:
@drarmstr  I noticed that even after your fix was released I still had a case of an effect setting an initial value that was having its value get lost. It seemed related to this bug, so I'm mentioning it here. I opened PR #991 to demonstrate the incorrect behavior. 

I spent a little time debugging the code to look for a fix based on your previous fix, but this seems different enough that a new fix didn't jump out at me. I hope my testcase can help somebody debug the problem. If you point me at a rough place that might be causing the problem, I'm happy to take another stab at fixing it myself

drarmstr:
@mdlavin - Thank you very much for the test case!  Yes, I suspect this test is unrelated to the original issue here.

PR #991, opened 2021-04-24
Failing testcase for sync setSelf bug

mdlavin:
Even after the new 0.2.0 release including a fix for #774 I'm still seeing a problem with a synchronous setSelf call not taking effect immediately.  I've narrowed it down to a fairly small testcase that demonstrates the problem.

There are two code tweaks you can make to the testcase that will cause it to pass, and they might highlight where the underlying bug is:
1. If you remove the `async` modifier [here](https://github.com/facebookexperimental/Recoil/compare/master...mdlavin:sync-setSelf-after-async-dependency?expand=1#diff-1d5cf4e2e9b5f144d17cf87d8ec6ad1e46adea67bcf6a54303b7267c3ca01931R191) the test passes
2. If you flip the order of the dependencies [here](https://github.com/facebookexperimental/Recoil/compare/master...mdlavin:sync-setSelf-after-async-dependency?expand=1#diff-1d5cf4e2e9b5f144d17cf87d8ec6ad1e46adea67bcf6a54303b7267c3ca01931R192-R193), then the test passes

### DIFF
--- a/src/recoil_values/__tests__/Recoil_selector-test.js
+++ b/src/recoil_values/__tests__/Recoil_selector-test.js
@@ -164,6 +164,46 @@ testRecoil('selector reset', () => {
   expect(get(selectorRW)).toEqual('DEFAULT');
 });
 
+testRecoil(
+  'selector get - sync setSelf dependency after async dependency',
+  async () => {
+    const syncSetSelfAtom = atom({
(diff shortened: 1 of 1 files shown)


drarmstr:
@csantos42 - This this is related to #998?

PR #1001, opened 2021-04-29
Optional setter parameter

Berkays:
I wanted to able to use optional parameters for selector setter function but type definitions does not allow this. See #1000 for my use case.


Berkays:
No need to merge as it can be solved as in #1000

PR #1024, opened 2021-05-14
Fix newValue option in ReadWriteSelectorOptions<T>

EdwardDrapkin:
Currently, the type for `newValue` is specified as `T | DefaultValue`; however, T is defined as a tagging interface:

```tsx
export class DefaultValue {
  private __tag: 'DefaultValue';
}
```

This effectively forces consumers to do unnecessary type refinement at the usage site because TS thinks that the DefaultValue is _just_ the tag, and not the data as well.

This proposed change will address this particular case.

In order to address this anywhere else it occurs, perhaps DefaultValue should be redefined as:

```tsx
export type DefaultValue<T> = T & {
  private __tag: 'DefaultValue';
}
```

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -122,7 +122,7 @@ export interface ReadWriteSelectorOptions<T> extends ReadOnlySelectorOptions<T> 
       get: GetRecoilValue;
       reset: ResetRecoilState;
     },
-    newValue: T | DefaultValue,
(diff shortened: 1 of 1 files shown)


drarmstr:
Hmm, I'm not quite following how `T & DefaultValue` is actually valid.  Could you provide some usage examples?  Note that the user often does need to do type-refinement when setting a new selector value in order to properly handle the "reset" case.

PR #1066, opened 2021-06-11
[docs] Document "Scoped" Atoms

drarmstr:


docs/docs/api-reference/core/atom.md
```diff
@@ -63,3 +63,7 @@ function Counter() {
   );
 }
 ```
+
+## Scoped Atoms
```
davidmccabe, line 67: I would title this section "Atom Families" and say "scoping your atom state or storing a collection of related states." or something like that.

docs/docs/api-reference/utils/atomFamily.md
```diff
@@ -64,7 +64,7 @@ const myAtomFamily = atomFamily({
 });
 ```
 
-or using [`selectorFamily`](/docs/api-reference/utils/selectorFamily) instead of `selector`, you can also access the parameter value in a `default` selector as well.
+or using [`selectorFamily()`](/docs/api-reference/utils/selectorFamily) you can have a dynamic `default` with access to the parameter value.  Don't just use `selector()` for `atomFamily()` defaults, as it would produce duplicate keys.
```
davidmccabe, line 67: I don't understand this.

davidmccabe (approved):
Stamped to unblock but please see the comments.

docs/docs/api-reference/utils/atomFamily.md
drarmstr: wordsmithed a bit more

PR #1109, opened 2021-07-09
Update waitForAny.md

ElRatonDeFuego:
"until [...] one [...] are" => "until [...] one [...] is"


drarmstr:
@mondaychen - I believe in this case "is" is actually correct since it is referencing "at least one"?
 https://www.reddit.com/r/grammar/comments/j5erm1/which_of_these_sentences_are_correct/ (Reddit obviously isn't the most reputable cite..)

mondaychen:
@drarmstr yes, this PR is changing "are" to "is"

PR #1116, opened 2021-07-13
Fix TS types for getNodes and useGetRecoilValueInfo

AssafKr:


bryan-hunter (approved):
I see you @AssafKr :)

AssafKr:
Hey @bryan-hunter! 👋

drarmstr:
Thanks! Looks like `tests.ts` will also have to be updated.

drarmstr:
Landed with #1123

PR #1123, opened 2021-07-22
Fix up some Typescript

drarmstr:
Summary: Fix up some TypeScript exports.  Based on #1116 from @AssafKr and fix up tests.  Also drive-by some changes so we work with both TypeScript 3.8 and 3.9 which had some minor changes in formatting and which lines errors show up on.

Differential Revision: D29863364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ## NEXT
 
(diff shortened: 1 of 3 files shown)


drarmstr:
landed

PR #1141, opened 2021-07-29
Documentation for 0.4

drarmstr:
* Release notes
* `useRecoilTransaction_UNSTABLE()`
* Breaking changes for Atom Effect `onSet()`
* `cachePolicy_UNSTABLE` - See also #1139

### DIFF
--- a/docs/blog/2021-07-30-0.4.0-release.md
+++ b/docs/blog/2021-07-30-0.4.0-release.md
@@ -6,7 +6,7 @@ We are pleased to announce the release of Recoil 0.4 with initial Recoil Transac
 
 ## Recoil Transactions
 
-Introducing Recoil Transactions!  The new [`useRecoilTransaction_UNSTABLE()`](/docs/api-reference/core/useRecoilTransaction) API allow for an easier, more performant, and safer way to atomically update multiple atoms together.  For the simple case this could be thought of as similar to the "updater" form of setting Recoil state, but over multiple atoms.  This new hook is easier to use and should replace the need to use the [`useRecoilCallback()`](/docs/api-reference/core/useRecoilCallback) API most of the time.
+Introducing Recoil Transactions!  The new [`useRecoilTransaction_UNSTABLE()`](/docs/api-reference/core/useRecoilTransaction) API allows for an easier, more performant, and safer way to atomically update multiple atoms together.  For the simple case this could be thought of as similar to the "updater" form of setting Recoil state, but over multiple atoms.  This new hook is easier to use and should replace the need to use the [`useRecoilCallback()`](/docs/api-reference/core/useRecoilCallback) API most of the time.
 
 *This is the initial implementation of the hook.  There are various limitations and future plans for transactions, please [read the docs](/docs/api-reference/core/useRecoilTransaction#current-limitations-and-future-vision).*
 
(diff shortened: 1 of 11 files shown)


docs/blog/2021-07-30-0.4.0-release.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Recoil 0.4
+---
+
+We are pleased to announce the release of Recoil 0.4 with initial Recoil Transactions, selector cache configuration, and other optimizations and fixes.
```
davidmccabe, line 5: "initial Recoil Transactions" doesn't make sense without further context. I would say "an improved API for updating multiple atoms together". So:

We are pleased to announce Recoil 0.4, with configurable selector caches, an improved API for updating multiple atoms together, and other improvements and fixes.

docs/blog/2021-07-30-0.4.0-release.md
```diff
@@ -0,0 +1,87 @@
...
+      break;
+
+    case 'turn':
+      set(headingState, action.heading);
+      break;
+  }
+});
+```
+
+## Selector Cache Policy Configuration
```
davidmccabe, line 72: I'd put this first, it's easier to explain (thus shorter) and probably of interest to more people.

docs/blog/2021-07-30-0.4.0-release.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Recoil 0.4
+---
+
+We are pleased to announce the release of Recoil 0.4 with initial Recoil Transactions, selector cache configuration, and other optimizations and fixes.
+
+## Recoil Transactions
```
davidmccabe, line 7: I am dubious that we should brand this as Recoil Transactions™ any more than the other thing should be Recoil Caches™. How about just 

## Updating Multiple Atoms

docs/blog/2021-07-30-0.4.0-release.md
```diff
@@ -0,0 +1,87 @@
...
+
+Introducing Recoil Transactions!  The new [`useRecoilTransaction_UNSTABLE()`](/docs/api-reference/core/useRecoilTransaction) API allows for an easier, more performant, and safer way to atomically update multiple atoms together.  For the simple case this could be thought of as similar to the "updater" form of setting Recoil state, but over multiple atoms.  This new hook is easier to use and should replace the need to use the [`useRecoilCallback()`](/docs/api-reference/core/useRecoilCallback) API most of the time.
```
davidmccabe, line 9: We're introducing a [new API] that lets you update multiple atoms together, as a single transaction, in a way that is easier, more efficient, and safer than before. This API replaces most uses of `useRecoilCallback`. However, this release contains only an initial implementation, with certain limitations that will be removed in a future release.

davidmccabe (changes requested):

docs/blog/2021-07-30-0.4.0-release.md
drarmstr: Yeah, good idea, was just thinking that also!

docs/blog/2021-07-30-0.4.0-release.md
drarmstr: I think the terminology and concept of "transactions" explains this well.  But yeah, the intention wasn't to brand it per-se.  Let me wordsmith it a bit and fix the capitalization, that shouldn't have been capitalized there..

PR #1177, opened 2021-08-20
setItem newValue FIX

cotapon:
If `newValue` is a string, double quotation marks will be added before and after the value, so only'typeof object' is modified to` json.stringify`.


mondaychen:
But when we read the value, we do apply `JSON.parse` to it

PR #1198, opened 2021-08-31
Add atoms/selectors state observer for testing.

timbergus:
I have added a description of an adaptation of the component described in the documentation in the developer tools section to be used in testing. This is a simple observer that receives an atom/selector and a function called every time it changes its state. So you can observe the function behavior to evaluate the result of the user's actions.

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -24,3 +24,78 @@ test('Test multipliedState', () => {
   expect(testSnapshot.getLoadable(multipliedState).valueOrThrow()).toBe(100);
 })
 ```
+
+## Testing Recoil Selectors inside of a React Component
+
+It can be useful to know the state inside a component when testing it. You can use the state to evaluate the user's actions, comparing it with an expected state. As described [`here`](/docs/guides/dev-tools#observing-all-state-changes), you can use this component designed for the developer tools, also for testing your code.
+
+```jsx
+export const RecoilObserver = ({ element, onChange } => {
+  const value = useRecoilValue(element)
+  useEffect(() => onChange(value), [onChange, value])
+  return null
+}
+```
+
+* Element: can be an atom or a selector.
+* onChange: this function will be called everytime the state changes.
+
+### Example: Form state modified by user
+
+#### Component
+
+```jsx
+import { atom, useRecoilState } from 'recoil';
+
+export const nameAtom = atom({
+  key: 'nameAtom',
+  default: '',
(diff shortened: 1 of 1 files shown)


docs/docs/guides/testing.md
```diff
@@ -24,3 +24,78 @@ test('Test multipliedState', () => {
   expect(testSnapshot.getLoadable(multipliedState).valueOrThrow()).toBe(100);
 })
 ```
+
+## Testing Recoil Selectors inside of a React Component
+
+It can be useful to know the state inside a component when testing it. You can use the state to evaluate the user's actions, comparing it with an expected state. As described [`here`](/docs/guides/dev-tools#observing-all-state-changes), you can use this component designed for the developer tools, also for testing your code.
```
drarmstr, line 30: That section in the guide on dev tools focuses on getting a Snapshot for all state.  That can be useful, and perhaps worth another example, but your example below just uses the normal `useRecoilValue()` hook.  This is great and simple!  But, I guess keep the reference to the dev tools guide in a separate section from this example then.

docs/docs/guides/testing.md
```diff
@@ -24,3 +24,78 @@ test('Test multipliedState', () => {
   expect(testSnapshot.getLoadable(multipliedState).valueOrThrow()).toBe(100);
 })
 ```
+
+## Testing Recoil Selectors inside of a React Component
```
drarmstr, line 28: Your example is probably a better one to start with.  Let's start with it as the beginning of the guide!

drarmstr (commented):
Thank you so much for helping to improve the documentation!!

docs/docs/guides/testing.md
timbergus: I can remove the reference from here. Maybe mixing `useRecoilValue` with the snapshot mechanics can confuse people.

docs/docs/guides/testing.md
timbergus: OK. Will do. Thanks a lot 😄

timbergus:
> Thank you so much for helping to improve the documentation!!

My pleasure 🙂

drarmstr:
Thanks again @timbergus  for helping with the docs!!

timbergus:
> Thanks again @timbergus for helping with the docs!!

Any time @drarmstr. I'm glad that I was able to contribute to this project. Thanks a lot for your effort to bring it to us. It's awesome!

PR #1204, opened 2021-09-07
[docs] Testing guide polish

drarmstr:
Minor polish for new guide on testing from @timbergus #1198

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -77,7 +77,7 @@ describe('The form state should', () => {
 });
 ```
 
-## Testing Recoil State outside of React
+## Testing Recoil state outside of React
 
 It can be useful to manipulate and evaluate Recoil selectors outside of a React context for testing.  This can be done by working with a Recoil [`Snapshot`](/docs/api-reference/core/Snapshot).  You can build a fresh snapshot using `snapshot_UNSTABLE()` and then use that `Snapshot` to evaluate selectors for testing.
 
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -2,27 +2,27 @@
 title: Testing
 ---
 
-## Testing Recoil Selectors inside of a React Component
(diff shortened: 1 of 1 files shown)


docs/docs/guides/testing.md
```diff
@@ -2,27 +2,27 @@
 title: Testing
 ---
 
-## Testing Recoil Selectors inside of a React Component
+## Testing Recoil state inside of a React component
```
timbergus, line 5: Just for consistency, `state` in the second block is capitalized (`State`). Shouldn't be this one capitalized too? Or the other one in lower case.

docs/docs/guides/testing.md
```diff
@@ -49,26 +49,26 @@ export default Form;
 #### Test
 
 ```jsx
-import { RecoilRoot } from 'recoil';
-import { fireEvent, render, screen } from '@testing-library/react';
+import {RecoilRoot} from 'recoil';
```
timbergus, line 52: Just out of curiosity. Why not adding spaces around objects? I knew the rule about not leaving spaces around array contents, but I thought leaving spaces between object contents and the curly braces was acceptable.

timbergus (commented):
Thanks a lot for the review! It has more sense in talking about `nodes` than `elements`. And thanks for rephrasing the first paragraph. English is not my mother tongue, and sometimes I write in English thinking in Spanish 😅

docs/docs/guides/testing.md
drarmstr: thanks for the catch!

docs/docs/guides/testing.md
drarmstr: Yeah, spaces there are pretty common with open source projects.  However, the internal Facebook prettier standard has no spaces.  So, this change is just for consistency with the rest of the code base and samples.  No value judgement either way.

timbergus: Thanks. I didn't know 🙂

PR #1218, opened 2021-09-16
Fix  onSet param type

linguofeng:
`newValue` call by `useResetRecoilState` is `DefaultValue` type

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -89,7 +89,7 @@ export type AtomEffect<T> = (param: {
   // Subscribe callbacks to events.
   // Atom effect observers are called before global transaction observers
   onSet: (
-    param: (newValue: T, oldValue: T | DefaultValue) => void,
+    param: (newValue: T | DefaultValue, oldValue: T | DefaultValue) => void,
   ) => void,
 }) => void | (() => void);
 


drarmstr:
Do you have an example where `DefaultValue` is actually being provided for this parameter?  The interface was updated so that it should provide the actual default value itself in this parameter when the atom is reset instead of the `DefaultValue` placeholder.

linguofeng:
> Do you have an example where `DefaultValue` is actually being provided for this parameter? The interface was updated so that it should provide the actual default value itself in this parameter when the atom is reset instead of the `DefaultValue` placeholder.

```ts
export const TOKEN_KEY = 'token-key';

export const tokenState = atom<string>({
  key: 'tokenState',
  default: '',
  effects_UNSTABLE: [
    ({ setSelf, onSet }) => {
      const token = localStorage.getItem(TOKEN_KEY);
      if (token) {
        setSelf(JSON.parse(token));
      }
-     onSet(token => { // token is string type, The left-hand side of an 'instanceof' expression must be of type 'any', an object type or a type parameter.ts(2358)
+     onSet((token: string | DefaultValue) => { // is OK
        if (token instanceof DefaultValue) {
          localStorage.removeItem(TOKEN_KEY);
        } else {
          localStorage.setItem(TOKEN_KEY, JSON.stringify(token));
        }
      });
    },
  ],
});

```

linguofeng:
ref https://github.com/facebookexperimental/Recoil/blob/85f62d75c7250df84e27ad37acf374754e1116c5/typescript/index.d.ts#L365

drarmstr:
With #1059 this check is no longer meaningful as `DefaultValue` is not used for the new value as the actual default value is used. 
```jsx
        if (token instanceof DefaultValue) {
          localStorage.removeItem(TOKEN_KEY);
```
The current interface doesn't distinguish between observing resetting an atom and setting an atom to an equivalent value.  If you would like to make that distinction you could use `getInfo_UNSTABLE()` in the effect from #1210 (not yet landed).  Open to ideas for a more elegant approach.

PR #1303, opened 2021-10-18
docs: fix "fallback" translation

kongmoumou:
"回调" means callback, "回退" means fallback.


mondaychen:
Thanks for pointing this out!
While 回调 is apparently wrong, I feel 回退 is not that commonly accepted translation for fallback, and may cause confusion. How about 后备？We can also leave the original English word in brackets i.e. 后备（fallback）

PR #1444, opened 2021-11-19
[recoil] Fix React strict mode

drarmstr:
Fix Recoil for React strict mode which runs effects multiple times:
   * Now atoms are re-initialized after the effect cleanup handler has cleaned them up.
   * The Batcher will reenable the notification mechanism after it was disabled on unmounting to avoid late async updates from trying to re-render.
   * Audit all other useEffect() calls except for useRecoilInterface() and dev tools <Connector>

Inspired from D24722151

Differential Revision: D32572832

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## UPCOMING
 ***Add new changes here as they land***
 
+- Fix Recoil with React strict mode that re-executes effects multiple times.
 - Add `refresh` to Recoil callback interface (#1413)
 - Fix transitive selector refresh for some cases (#1409)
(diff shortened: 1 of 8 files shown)


MRHarrison:
@drarmstr what happened here, was this merged, closed or pushed out, I have issues with strict mode

drarmstr:
Yes, this was merged; it just wasn't released in a new NPM package yet.  You can use the nightly branch until then.

MRHarrison:
@drarmstr couldn't get the nightly to install, I guess I'll have to wait for the release.  I'm essentially getting a screen flash/reload when recoil initially loads.

Issue #1457, opened 2021-11-23
Cannot mock atom with asynchronous default function


NS-BOBBY-C:
Hi Recoil Team!

We have successfully integrated recoil into one of our applications and we have seen incredible improvements in app performance due to this library.

I am struggling with how to write jest tests for atoms which have an asynchronous promise as the default value. I've included a codesandbox with my struggle:

https://codesandbox.io/s/sleepy-lederberg-xg38e?file=/src/components/RecoilComponent/RecoilComponent.test.tsx

The pattern below is a pretty common pattern in our application, and I am struggling to test the components where we use recoil. I have a wrapper component in a custom render function, but I'm not sure how to wait for the fetchData function to return - and wrapping in act/waitFor doesn't seem to wait for the function to finish fetching.
 
```typescript
const fetchData = async (id: string) => fetch("API/path/here")

export const dataFromId = memoize((id: string) => atom({ key: 'Atom_${id}, default: fetchData(id) })
```

drarmstr:
In our unit testing we have a function: 
```jsx
function flushPromisesAndTimers(): Promise<void> {
  return act(
    () =>
      new Promise(resolve => {
        setTimeout(resolve, 100);
        jest.runAllTimers();
      }),
  );
}
```
Then in the tests you can do:
```jsx
  await flushPromisesAndTimers();
```
to resolve them.  If this helps you, please let us know and we can update [the documentation](https://recoiljs.org/docs/guides/testing).

Also, you may wish to use `atomFamily()` instead of direct memoization of atoms.

NS-BOBBY-C:
Thank you for the response! This does fix the issue for me.

I can update the documentation if you would like - I'll make a PR with an example.

For those of you trying this in codesandbox like I was - there is a lingering [issue](https://github.com/codesandbox/codesandbox-client/issues/513) that jest doesn't work 100% in codesandbox.

Side note:
What is the advantage of using atomFamily over memoization? Just that it keeps things in the recoil world?

PR #1463, opened 2021-11-24
Update testing.md with asynchronous default atom

NS-BOBBY-C:
Adds documentation for the discussion on #1457

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -22,15 +22,13 @@ export const RecoilObserver = ({node, onChange}) => {
 #### Component
 
 ```jsx
-import {atom, useRecoilState} from 'recoil';
-
-export const nameAtom = atom({
+const nameState = atom({
   key: 'nameAtom',
   default: '',
 });
 
 function Form() {
-  const [name, setName] = useRecoilState(nameAtom);
+  const [name, setName] = useRecoilState(nameState);
   return (
     <form>
       <input
@@ -42,19 +40,11 @@ function Form() {
     </form>
   );
 }
-
-export default Form;
 ```
 
 #### Test
 
 ```jsx
-import {RecoilRoot} from 'recoil';
-import {fireEvent, render, screen} from '@testing-library/react';
-
-import Form, {nameAtom} from './form';
-import {RecoilObserver} from './RecoilObserver';
-
 describe('The form state should', () => {
   test('change when the user enters a name.', () => {
     const onChange = jest.fn();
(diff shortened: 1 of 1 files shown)


docs/docs/guides/testing.md
```diff
@@ -77,6 +77,87 @@ describe('The form state should', () => {
 });
 ```
 
+## Testing Recoil state with asyncronous default inside of a React component
+
+A common pattern for the default state of atoms is to have an asynchronous function fetch the default state of an atom. However, while testing, the component is immediately suspended upon mounting and will not update in the dom without acting. To test this scenario, we need a helper function:
```
drarmstr, line 82: This pattern isn't just for async atom defaults, but it useful for testing any async dependency in Recoil, such as async selectors or effects as well.

docs/docs/guides/testing.md
```diff
@@ -77,6 +77,87 @@ describe('The form state should', () => {
...
+
+```jsx
+// testing-utils.js
+import {act} from '@testing-library/react'
```
drarmstr, line 86: We generally don't include the import statements for the documentation examples, particularly as they can depend on the user's environment.

docs/docs/guides/testing.md
```diff
@@ -77,6 +77,87 @@ describe('The form state should', () => {
...
+
+```jsx
+import {atom, useRecoilState} from 'recoil';
+
+const getDefaultTitleAtomState = async () => {
+  const response = await fetch('https://example.com/returns/a/json');
+  return await response.json(); // { title: 'real title' };
+}
+
+export const titleAtom = atom({
```
drarmstr, line 112: Naming convention for the documentation would be "titleState"

docs/docs/guides/testing.md
```diff
@@ -77,6 +77,87 @@ describe('The form state should', () => {
...
+function Title() {
+  const data = useRecoilValue(titleAtom);
+  return (
+    <div>
+      <h1>{data.title}</h1>
+    </div>
+  );
+}
+
+export default Title;
```
drarmstr, line 126: Don't both with the formal "export" statements for documentation either, same as the imports.

docs/docs/guides/testing.md
```diff
@@ -77,6 +77,87 @@ describe('The form state should', () => {
...
+      <RecoilRoot>
+        <Suspense fallback={<div>loading...</div>}>
+          <Title />
+        </Suspense>
+      </RecoilRoot>
+    );
+    await flushPromisesAndTimers();
+    
+    expect(screen.getByText(mockState.title)).toBeInTheDocument()
+    expect(screen.getByText('Loading...')).not.toBeInTheDocument()
```
drarmstr, line 156: Consistent capitalization of "loading..."

drarmstr (changes requested):
Thanks for helping to improve the documentation!

docs/docs/guides/testing.md
NS-BOBBY-C: How should I best describe this in the summary? 

I was thinking something like:

> A common pattern for atoms is using asynchronous queries fetch the state of the atom, selector, or effect. This causes the component to be suspended. However, while testing, the component is immediately suspended upon mounting and will not update in the DOM without acting. To test this scenario, we need a helper function:

docs/docs/guides/testing.md
NS-BOBBY-C: OK. I included them in my example since the rest of the file included them. I can make things consistent throughout the file!

docs/docs/guides/testing.md
drarmstr: Or maybe `"It is common for atoms and selectors to have asynchronous dependencies or queries.  This can cause components to be suspended during initial rendering when testing.  When using Jest, you should use `act()` to wait for the promises to resolve and update the DOM.  Here is an example helper function:"`

drarmstr (approved):

PR #1469, opened 2021-11-30
Update atoms.mdx

franhhk:
The API reference specifies that the `useRecoilState()` hook should return a tuple of the current value of the state and a setter function. Thus, assuming that `setTodoList` is intended to be used as a setter function, the variable should be declared as the second value of the return tuple, instead of it being the tuple itself.

### DIFF
--- a/docs/i18n/ko/docusaurus-plugin-content-docs/current/basic-tutorial/atoms.mdx
+++ b/docs/i18n/ko/docusaurus-plugin-content-docs/current/basic-tutorial/atoms.mdx
@@ -40,7 +40,7 @@ function TodoList() {
 ```jsx
 function TodoItemCreator() {
   const [inputValue, setInputValue] = useState('');
-  const setTodoList = useSetRecoilState(todoListState);
+  const [todoList, setTodoList] = useSetRecoilState(todoListState);
 
   const addItem = () => {
     setTodoList((oldTodoList) => [


docs/i18n/ko/docusaurus-plugin-content-docs/current/basic-tutorial/atoms.mdx
```diff
@@ -40,7 +40,7 @@ function TodoList() {
 ```jsx
 function TodoItemCreator() {
   const [inputValue, setInputValue] = useState('');
-  const setTodoList = useSetRecoilState(todoListState);
+  const [todoList, setTodoList] = useSetRecoilState(todoListState);
```
drarmstr, line 43: Part of the point of this section is to demonstrate `useSetRecoilState()` which can be used to set an atom without reading it's state or subscribing the component to re-render when it changes.

drarmstr (changes requested):

docs/i18n/ko/docusaurus-plugin-content-docs/current/basic-tutorial/atoms.mdx
franhhk: I understood! I must have gotten confused with the `useRecoilState()` hook.

PR #1515, opened 2021-12-23
fix selectorGetter performance issue

thomaszdxsn:
- motivation: the getter is call `setDepsInStore` every time, harm the performance when multiple call the getter
- notes: this pr just a demo, I don't known is this feasible. please give me a confirm, and I will make some further change with testcases

#914 Please refer this issue for detail

### DIFF
--- a/packages/recoil/recoil_values/Recoil_selector.js
+++ b/packages/recoil/recoil_values/Recoil_selector.js
@@ -683,18 +683,6 @@ function selector<T>(
     }
   }
 
-  function setNewDepInStore(
-    store: Store,
-    state: TreeState,
-    deps: Set<NodeKey>,
-    newDepKey: NodeKey,
-    executionId: ?ExecutionId,
-  ): void {
-    deps.add(newDepKey);
-
-    setDepsInStore(store, state, deps, executionId);
-  }
-
   function evaluateSelectorGetter(
     store: Store,
     state: TreeState,
@@ -729,12 +717,10 @@ function selector<T>(
      */
     const deps = new Set();
 
-    setDepsInStore(store, state, deps, executionId);
-
     function getRecoilValue<S>(dep: RecoilValue<S>): S {
       const {key: depKey} = dep;
 
-      setNewDepInStore(store, state, deps, depKey, executionId);
+      deps.add(depKey);
 
       const depLoadable = getCachedNodeLoadable(store, state, depKey);
 
@@ -818,6 +804,7 @@ function selector<T>(
       maybeFreezeValue(loadable.contents);
     }
 
+    setDepsInStore(store, state, deps, executionId);
     return [loadable, depValues];
   }
 
--- a/packages/recoil/recoil_values/__tests__/Recoil_selector-test.js
+++ b/packages/recoil/recoil_values/__tests__/Recoil_selector-test.js
@@ -660,6 +660,52 @@ testRecoil(
   },
 );
 
+testRecoil('selector dep changed on async flying', async () => {
+  const resolvingSel1 = resolvingAsyncSelector(1);
+  const resolvingSel2 = resolvingAsyncSelector(2);
+  const anAtom3 = atom({key: 'atomTrackedAsync3', default: 3});
+
+  const selectorWithAsyncDeps = selector({
+    key: 'selectorDepChangeOnResolving',
+    get: async ({get}) => {
+      const val1 = get(resolvingSel1);
+
+      await Promise.resolve();
+
+      const val3 = get(anAtom3);
+
+      const val2 = get(resolvingSel2);
+      await Promise.resolve();
+
+      return val1 + val2 + val3;
+    },
+  });
+
+  let setAtom;
+
+  function Component() {
+    [, setAtom] = useRecoilState(anAtom3);
+    const selVal = useRecoilValue(selectorWithAsyncDeps);
+
+    return selVal;
+  }
+
+  const container = renderElements(<Component />);
+
+  expect(container.textContent).toEqual('loading');
+
+  await flushPromisesAndTimers();
+
+  act(() => setAtom(4));
+  // HACK: not sure why but these are needed in OSS
+  await flushPromisesAndTimers();
+  await flushPromisesAndTimers();
+  await flushPromisesAndTimers();
+  await flushPromisesAndTimers();
+
+  expect(container.textContent).toEqual('7');
+});
+
 testRecoil("Updating with same value doesn't rerender", ({gks}) => {
   if (!gks.includes('recoil_suppress_rerender_in_callback')) {
     return;
--- a/packages/recoil/recoil_values/__tests__/Recoil_selector-test.js
+++ b/packages/recoil/recoil_values/__tests__/Recoil_selector-test.js
@@ -1025,3 +1025,36 @@ testRecoil('Required options are provided when creating selectors', () => {
 
   window.__DEV__ = devStatus;
 });
+
+testRecoil('Dynamic deps discovered after the await was recorded', async () => {
+  const testSnapshot = freshSnapshot();
+  testSnapshot.retain();
+
+  const myAtom = atom({
+    key: 'my-atom',
+    default: 0,
+  });
+  let selectorRunCount = 0;
+  let selectorRunCompleteCount = 0;
+  const mySelector = selector({
+    key: 'my-selector',
+    get: async ({get}) => {
+      await Promise.resolve();
+      get(myAtom);
+      selectorRunCount++;
+      await new Promise(() => {});
+      selectorRunCompleteCount++;
+    },
+  });
+  testSnapshot.getLoadable(mySelector);
+  await flushPromisesAndTimers();
+  expect(selectorRunCount).toBe(1);
+  expect(selectorRunCompleteCount).toBe(0);
+  const mappedSnapshot = testSnapshot.map(({set}) =>
+    set(myAtom, prev => prev + 1),
+  );
+  expect(mappedSnapshot.getLoadable(mySelector).state === 'loading');
+  await flushPromisesAndTimers();
+  expect(selectorRunCount).toBe(2);
+  expect(selectorRunCompleteCount).toBe(0);
+});
--- a/packages/recoil/recoil_values/__tests__/Recoil_selector-test.js
+++ b/packages/recoil/recoil_values/__tests__/Recoil_selector-test.js
@@ -181,15 +181,6 @@ testRecoil('selector reset', () => {
   expect(getValue(selectorRW)).toEqual('DEFAULT');
 });
 
-testRecoil('useRecoilState - resolved async selector', async () => {
-  const resolvingSel = resolvingAsyncSelector('HELLO');
-  const c = renderElements(<ReadsAtom atom={resolvingSel} />);
-  expect(c.textContent).toEqual('loading');
-  act(() => jest.runAllTimers());
-  await flushPromisesAndTimers();
-  expect(c.textContent).toEqual('"HELLO"');
-});
-
 testRecoil('selector - evaluate to RecoilValue', () => {
   const atomA = atom({key: 'selector/const atom A', default: 'A'});
   const atomB = atom({key: 'selector/const atom B', default: 'B'});
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -1,11 +1,7 @@
 {
   "name": "recoil-shared",
   "description": "This is the internal package.json enabling CommonJS module",
-  "main": "Recoil_Shared.js",
   "haste_commonjs": true,
-  "files": [
-    "Recoil_Shared.js"
-  ],
   "directories": {
     "": "./"
   },


thomaszdxsn:
By the way, this change is fix my performance issue. But I don't know is it harm other issues?

thomaszdxsn:
@drarmstr give a review, please

drarmstr:
> @drarmstr give a review, please

Hi @thomaszdxsn !  Thank you so much for the contribution!  I'm still on a bit of a holiday schedule so I'll try to review later this week.  We do really appreciate your effort.

packages/recoil/recoil_values/Recoil_selector.js
```diff
@@ -700,7 +700,7 @@ function selector<T>(
   ): void {
     deps.add(newDepKey);
 
-    setDepsInStore(store, state, deps, executionId);
+    // setDepsInStore(store, state, deps, executionId);
```
drarmstr, line 703: With this simplification we can just get rid of `setNewDepInStore()` and call `deps.add()` directly in `getRecoilValue()`.

packages/recoil/recoil_values/Recoil_selector.js
```diff
@@ -782,6 +782,7 @@ function selector<T>(
 
     try {
       result = get({get: getRecoilValue, getCallback});
+      setDepsInStore(store, state, deps, executionId)
```
drarmstr, line 785: Best if this is done outside of this try/catch block.  We could just call it before the return, which references `depValues`, so it might be a nice logical place for it.

drarmstr:
Ok, I took a look; thanks again @thomaszdxsn !  I've actually thought an optimization like this makes sense, but we need to be careful with asynchronous cases. If a selector has pending dependencies, or is async itself, then it needs to update it's dependencies before relinquishing control; from a glance this may be done as there are calls to `setDepsInStore()` in the various promise wrapping logic.  The thing would be to have tests that confirm that a selector will properly be triggered to re-evaluate due to a dependency change if it is in the pending state either due to a pending dependency or due to itself still pending resolving its promise.  So, I think this is safe, but let's test it!

 @csantos42 do you have thoughts?

thomaszdxsn:
Ok, I will dive into internals

drarmstr:
Hi @thomaszdxsn, have you had a chance to dive in further?

thomaszdxsn:
Sorry, I am busy until 2022-01-26. after 01-25 I will give some further progress

drarmstr:
@thomaszdxsn - In the meantime, #1568 and #1566 do some unrelated cleanup of the selector implementation.  Hopefully it should help make it a bit more readable.

thomaszdxsn:
> @thomaszdxsn - In the meantime, #1568 and #1566 do some unrelated cleanup of the selector implementation. Hopefully it should help make it a bit more readable.

got it, thanks

thomaszdxsn:
Sorry, make some mess up so pr is closed.

The codebase is overwhelming for me.

>  from a glance this may be done as there are calls to setDepsInStore() in the various promise wrapping logic. The thing would be to have tests that confirm that a selector will properly be triggered to re-evaluate due to a dependency change if it is in the pending state either due to a pending dependency or due to itself still pending resolving its promise. So, I think this is safe, but let's test it!

- Do I need add code for "various promise wrapping log"?
- Do I need add more testcase? I do test already, and modify some resolvingSel with refresher, it make selector re-evaluate.
- Or I need add a new testcase, let the selector's deps change to variadic in runtime? like [a, b] -> [a, c, d]

drarmstr:
> Sorry, make a mess so pr is closed.
> 
> The codebase is overwhelming for me. please give some mentor
> 
> > from a glance this may be done as there are calls to setDepsInStore() in the various promise wrapping logic. The thing would be to have tests that confirm that a selector will properly be triggered to re-evaluate due to a dependency change if it is in the pending state either due to a pending dependency or due to itself still pending resolving its promise. So, I think this is safe, but let's test it!
> 
> * Do I need add code for "various promise wrapping log"?
> * Do I need add more testcase? I do test already, and modify some resolvingSel with refresher, it make selector re-evaluate.
> * Or I need add a new testcase, let the selector's deps change to variadic in runtime? like [a, b] -> [a, c, d]

Yeah, I'm hoping that no other promise wrapping logic changes are needed here.  But we should add more unit testcases in `Recoil_selector-test.js` or `Recoil_selectorHooks-test.js`.  At least something like:

Selector `await`'s on some `Promise`, which resolves, then `get()`'s a new dependency of some atom, then waits again on either another async dependency or `Promise` (test both).  Either way, the selector should still be in a pending state.  Then, we change the value of the atom and confirm the selector re-evaluates with the new value of the atom.  Based on the new value of the atom the selector can then just resolve and provide a value.  The key is that the dynamic dependency discovered after the `await` was recorded and changes to it cause the selector to re-evaluate.

thomaszdxsn:
Sorry I discover the new merged code now, Let me merge it

drarmstr:
@thomaszdxsn - Thanks for the updates so far.  I worked closer with your test case and updated it to cover some more cases.  

```jsx

  testRecoil('Dynamic deps will refresh', async () => {
    const myAtom = atom({
      key: 'selector dynamic deps atom',
      default: 'DEFAULT',
    });
    const myAtomA = atom({
      key: 'selector dynamic deps atom A',
      default: new Promise(() => {}),
    });
    const myAtomB = atom({
      key: 'selector dynamic deps atom B',
      default: 'B',
    });
    const myAtomC = atom({
      key: 'selector dynamic deps atom C',
      default: new Promise(() => {}),
    });

    let selectorEvaluations = 0;
    const mySelector = selector({
      key: 'selector dynamic deps selector',
      get: async ({get}) => {
        selectorEvaluations++;
        await Promise.resolve();
        const sw = get(myAtom);
        if (sw === 'A') {
          return 'RESOLVED_' + get(myAtomA);
        }
        if (sw === 'B') {
          return 'RESOLVED_' + get(myAtomB);
        }
        if (sw === 'C') {
          return 'RESOLVED_' + get(myAtomC);
        }
        await new Promise(() => {});
      },
    });

    const wrapperSelector = selector({
      key: 'selector dynamic deps wrapper',
      get: ({get}) => {
        const loadable = get(noWait(mySelector));
        return loadable.state === 'loading' ? 'loading' : loadable.contents;
      },
    });

    let setAtom, setAtomA, setAtomB;
    function ComponentSetter() {
      setAtom = useSetRecoilState(myAtom);
      setAtomA = useSetRecoilState(myAtomA);
      setAtomB = useSetRecoilState(myAtomB);
    }
    const c = renderElements(
      <>
        <ComponentSetter />
        <ReadsAtom atom={wrapperSelector} />
      </>,
    );
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(1);

    // Cause re-evaluation to pending state
    act(() => setAtom('TMP'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(2);

    // Add atomA dependency, which is pending
    act(() => setAtom('A'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(3);

    // change to atomB dependency
    act(() => setAtom('B'));
    await flushPromisesAndTimers();
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_B"');
    expect(selectorEvaluations).toBe(4);

    // Set atomB
    act(() => setAtomB('SETB-0'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_SETB-0"');
    expect(selectorEvaluations).toBe(5);

    // Change back to atomA dependency
    act(() => setAtom('A'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(6);

    // Setting B is currently ignored
    act(() => setAtomB('SETB-IGNORE'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(6);

    // Set atomA
    act(() => setAtomA('SETA'));
    await flushPromisesAndTimers();
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_SETA"');
    expect(selectorEvaluations).toBe(7);

    // Setting atomB is ignored
    act(() => setAtomB('SETB-LATER'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_SETA"');
    expect(selectorEvaluations).toBe(7);

    // Change to atomC, which is pending
    act(() => setAtom('C'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(8);

    // Setting atomA is ignored
    act(() => setAtomA('SETA-LATER'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"loading"');
    expect(selectorEvaluations).toBe(8);

    // change back to atomA for new value
    act(() => setAtom('A'));
    await flushPromisesAndTimers();
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_SETA-LATER"');
    expect(selectorEvaluations).toBe(9);

    // Change back to atomB
    act(() => setAtom('B'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_SETB-LATER"');
    expect(selectorEvaluations).toBe(10);

    // Set atomB
    act(() => setAtomB('SETB-1'));
    await flushPromisesAndTimers();
    expect(c.textContent).toBe('"RESOLVED_SETB-1"');
    expect(selectorEvaluations).toBe(11);
  });
```

You'll notice this also adds a wrapper selector using `noWait()`.  It turns out that using a snapshot or just the selector directly causes the component using it to suspend and thus re-evaluate on wake even if the dependencies aren't sufficient.  Thee wrapper helps avoid this and reveals this approach isn't working.  Restarting the evaluation is reseting the deps list and  the partial evaluation promises are never resolving so they don't have a chance to set them either.  So, I don't think we can avoid updating deps when calling `getRecoilValue()`...  

But, I still think there is hope here if we can just optimize `saveDependencyMapToStore()`.  For example, maybe updating [`mergeDependencyMapIntoGraph()`](https://github.com/facebookexperimental/Recoil/blob/main/packages/recoil/core/Recoil_Graph.js#L43) to only go through the expensive update if thee graph's dependencies don't already match what it is trying to set it to.  So, perhaps it could first check if the work is already done and then abort.  What do you think?

thomaszdxsn:
Involve async things make it very complicate. actually I don't use any async in recoil(I have a little async logic, but write as `redux-thunk` style. etc async then update atom).So I don't have use many API in Recoil like `noWait`.

I known what you meaning but don't know how to optimize `saveDependencyMapToStore`

thomaszdxsn:
My scenario is:

1. A atomFamily representation a linked list nodes
2. UI as a Tree
3. Every atom has a collapse/expand field
4. A Selector calculate display nodes based on collapse and children
5. Each time user click the collapse button, the selector is reevaluate.

When display nodes show more than 900+, the response time is 500ms+. Because every `get` trigger `mergeDependencyMapIntoGraph()` is very expensive

thomaszdxsn:
> For example, maybe updating [mergeDependencyMapIntoGraph()](https://github.com/facebookexperimental/Recoil/blob/main/packages/recoil/core/Recoil_Graph.js#L43) to only go through the expensive update if thee graph's dependencies don't already match what it is trying to set it to. So, perhaps it could first check if the work is already done and then abort...

What you mean `if thee graph's dependencies don't already match what it is trying to set it to`?  does this API can call only once?

drarmstr:
Yeah, even though many have simple use cases we still need to be careful to make sure all the corner cases work for everyone.

The possible optimization is perhaps just checking if the graph needs to be updated before updating it.  Thinking is that the updating process, with taking multiple set differences, is more expensive than just first checking if they are already equivalent up-front..

thomaszdxsn:
Sorry but what is timing for `dependencies don't already match`? for my view the deps don't match after every `get()` called

thomaszdxsn:
> Sorry but what is timing for `dependencies don't already match`? for my view the deps don't match after every `get()` called

drarmstr:
> @drarmstr Sorry but what is timing for `dependencies don't already match`? for my view the deps don't match after every `get()` called

Ah, I see, you're trying to cover the case of looping over new dependencies, not just existing dependencies.  Yeah, then that idea won't be sufficient either.  

Hmm, I guess it's worth diving into the actual graph update algorithm in `Recoil_Graph.js`, then, as that could likely be heavily optimized...

drarmstr:
Add better unit test for dynamic dependencies in #1598

drarmstr:
@thomaszdxsn - An idea to optimize the common case, while still supporting all cases, might be to lazily update all of the dependencies that are synchronously requested by the selector from the initial execution of the `get()` callback, then handle the remaining async dependencies on-demand as now.  Hopefully that should help cover your situation and many others while still protecting correctness.

thomaszdxsn:
> An idea to optimize the common case, while still supporting all cases, might be to lazily update all of the dependencies that are synchronously requested by the selector from the initial execution of the `get()` callback, then handle the remaining async dependencies on-demand as now. Hopefully that should help cover your situation and many others while still protecting correctness.

Can you give some pseudo code?

drarmstr:
> > An idea to optimize the common case, while still supporting all cases, might be to lazily update all of the dependencies that are synchronously requested by the selector from the initial execution of the `get()` callback, then handle the remaining async dependencies on-demand as now. Hopefully that should help cover your situation and many others while still protecting correctness.
> 
> Can you give some pseudo code?

It might be as easy:
* Set a boolean as true at the top of `evaluateSelectorGetter()`
* Call `setDepsInStore()` at the end of `evaluateSelectorGetter()`, as you added and set that boolean as false
* Then, change `getRecoilValue()` to conditionally call `setDepsInStore()` based on the boolean being false.

drarmstr:
@thomaszdxsn - Ok, I've got it implemented and it seems to work really well.  You don't have to worry about updating the PR, I can take care of merging it.  Thanks for all of your help here, this will be a great improvement!

thomaszdxsn:
> @thomaszdxsn - Ok, I've got it implemented and it seems to work really well. You don't have to worry about updating the PR, I can take care of merging it. Thanks for all of your help here, this will be a great improvement!
> 
> (Also note #1651)

You are very gorgeous man, thank you very much, I learn a lot of things from this issue

PR #1520, opened 2021-12-31
Rename effects_UNSTABLE to effects

drarmstr:
Summary: For Atom effects rename `effects_UNSTABLE` option to just `effects` as the interface has mostly stabalized.  Continue to support `effects_UNSTBLE` for backward compatibility for now.

Differential Revision: D33373696


thomaszdxsn:
Are you finish this in one day? It shock me!

Issue #1522, opened 2022-01-03
Include testing example using react-hooks-testing-library


patrickspafford:
Hi,

While the [Recoil testing documentation](https://recoiljs.org/docs/guides/testing) is helpful, I think it could be improved if there was mention / an example of how to test with React Hooks Testing Library. Because atoms and selectors require nesting inside RecoilRoot, the example could show something along these lines (see [Advanced Hooks testing](https://react-hooks-testing-library.com/usage/advanced-hooks) for where this comes from):

```javascript
// state.ts
const countAtom = atom({
key: 'countAtom',
default: 0
})

// useMyCustomHook.ts
const useMyCustomHook = () => {
  const [count, setCount] = useRecoilState(countAtom)
  // insert other state here
  // insert your custom useEffects here (that are not best written as atom effects)
  return [count, setCount]
}

// useMyCustomHook.test.ts
test('My test', () => {
  const { result } = renderHook(() => useMyCustomHook(), {
    wrapper: RecoilRoot, // <-- this is the important part. We can now access Recoil values because they are properly wrapped
  })
 expect(result.current).toEqual(0)
})
```

With this approach, I'm not sure that you'd need to use snapshots and this way seems less cumbersome. It seems to work for me in a React Native project anyway...

Thanks so much

drarmstr:
Thanks, feel free to make a PR with the doc improvements!  Note that for our internal testing we also always wrap the tests with suspense and error boundaries in addition to `<RecoilRoot>`.

patrickspafford:
If you can point me in the direction of the docs, will do. But it looks like the docs are included in the .gitignore?

drarmstr:
Thanks again @patrickspafford !

PR #1528, opened 2022-01-06
[docs] Document custom hooks that use Recoil states

patrickspafford:
Per #1522, include an example of how to test a custom hook that relies on Recoil states. This example uses React Hooks Testing Library

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -4,7 +4,7 @@ title: Testing
 
 ## Testing Recoil state inside of a React component
 
-It can be helpful to test Recoil state when testing a component. You can compare the new state  against expected values using this pattern. It uses a React functional component, `useRecoilValue` and `useEffect`, to observe an `atom`/`selector`'s changes and execute a callback every time the user performs an action that modifies the state.
+It can be helpful to test Recoil state when testing a component. You can compare the new state against expected values using this pattern. It uses a React functional component, `useRecoilValue` and `useEffect`, to observe an `atom`/`selector`'s changes and execute a callback every time the user performs an action that modifies the state.
 
 ```jsx
 export const RecoilObserver = ({node, onChange}) => {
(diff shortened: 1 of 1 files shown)


docs/docs/guides/testing.md
```diff
@@ -76,7 +76,7 @@ A common pattern for atoms is using asynchronous queries fetch the state of the
 function flushPromisesAndTimers(): Promise<void> {
   return act(
     () =>
-      new Promise(resolve => {
+      new Promise((resolve) => {
```
drarmstr, line 79: For consistency, our prettier configuration does not wrap single parameters with parens.

docs/docs/guides/testing.md
```diff
@@ -114,31 +114,71 @@ function Title() {
...
+
+    expect(screen.getByText(mockState.title)).toBeInTheDocument();
+    expect(screen.getByText('loading...')).not.toBeInTheDocument();
+  });
+});
+```
+
+## Testing Recoil state inside a Custom Hook
+
+Sometimes it becomes convenient to write a custom React hook that relies on Recoil states. Since you test the hook in a React context, it also needs to be wrapped in `RecoilRoot`. Luckily, you can use [React Hooks Testing Library](https://react-hooks-testing-library.com/) to achieve this.
```
drarmstr, line 141: Sometimes it is convenient to write custom React hooks that rely on Recoil state.  These need to be wrapped in a `<RecoilRoot>`.  The [React Hooks Testing Library](https://react-hooks-testing-library.com/) can help with this pattern.

docs/docs/guides/testing.md
```diff
@@ -114,31 +114,71 @@ function Title() {
...
+});
+```
+
+#### Hook
+
+```jsx
+const useMyCustomHook = () => {
+  const [count, setCount] = useRecoilState(countState);
+  // Insert other Recoil state here...
+  // Insert other effects here...
```
drarmstr, line 160: Maybe just "Insert other hook logic here...", which may or may not be an "effect"

docs/docs/guides/testing.md
```diff
@@ -114,31 +114,71 @@ function Title() {
...
+  return {
+    count,
+    setCount,
+  };
```
drarmstr, line 164:
```
return { count, setCount };
```

If you're going to include `setCount` here, then the example should also demonstrate using it as well.

drarmstr (changes requested):

drarmstr (approved):

PR #1577, opened 2022-01-29
Fix postInstall script

drarmstr:
Summary: D32366885 (https://github.com/facebookexperimental/recoil/commit/97cb9783400f18c22797e822af998fbf1cfca244) added a `postInstall` script for building the temporary patched package we use for `gen-flow-files`.  However, this was added as a production install script when the `devDependencies` are not available.  It should only be used at build time so move to `postbuild.js` script.

Differential Revision: D33871387

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "recoil",
-  "version": "0.6.0",
+  "version": "0.6.1",
   "description": "Recoil - A state management library for React",
   "main": "cjs/recoil.js",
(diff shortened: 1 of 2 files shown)


micahbule:
This broke one of our projects. Waiting for the fix. Thanks for this.

drarmstr:
> This broke one of our projects. Waiting for the fix. Thanks for this.

Fix is in 0.6.1

PR #1592, opened 2022-02-07
Add note about retaining a snapshot in async selector jest tests

### LINKED ISSUE #1590: Atom with async selector not working in jest tests after upgrading from 0.5 to 0.6
Hi,

I'm using an [atom](https://github.com/vntw/recoil-atom-async-selector/blob/f7351cb6241cc95a0cfe8d0d3be8785f510adb1a/src/recoil/atoms.ts#L9-L21) with an async selector and a jest [test](https://github.com/vntw/recoil-atom-async-selector/blob/f7351cb6241cc95a0cfe8d0d3be8785f510adb1a/src/recoil/atoms.test.ts#L7-L24). These tests work in recoil 0.5 but fail with 0.6 with the following errors:

```
// 1st test
thrown: Canceled {}

// 2nd test
Received promise rejected instead of resolved
Rejected to value: {}
```

Reading [these](https://recoiljs.org/docs/guides/testing#example-jest-unit-testing-selectors) and [these](https://recoiljs.org/docs/guides/asynchronous-data-queries/#query-default-atom-values) docs it still _seems_ to be correct or at least in that direction. The release notes mention some changes around Canceled, this is maybe related?

To reproduce, I created two branches:
- Checkout and install deps
  - Branch `0.5.2` (working): https://github.com/vntw/recoil-atom-async-selector/tree/0.5.2
  - Branch `0.6.1` (broken): https://github.com/vntw/recoil-atom-async-selector/tree/0.6.1
- Run `yarn test`, a second test is commented out to avoid cached selector issues but these create a different error (maybe because of wrong usage)

Would be great to hear if this is a regression and if not, how these type of atoms/tests should be written now instead since I couldn't find it in the docs (exactly)

Thanks!

vntw:
I wasn't sure if it's necessary to create a complete second example since they're _mostly_ equal in code. Right now it only references the already existing documentation but if you prefer an example with a new headline just let me know, I originally prepared one:

```js
export const numberState = atom({
  key: "number",
  default: selector({
    key: "number/Default",
    get: async () => {
      // await some numbers
    }
  })
});

test('Test async numberState default', async () => {
  const initialSnapshot = snapshot_UNSTABLE();
  initialSnapshot.retain();
  await expect(initialSnapshot.getLoadable(numberState).promiseOrThrow()).resolves.toBe(0);
});
```

This could also link to the existing docs for a more complete picture of retain/release handling

Closes #1590

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -196,6 +196,15 @@ test('Test multipliedState', () => {
 });
 ```
 
+### Jest unit testing async selectors
+
+When testing an **async selector**, it is necessary to `retain()` the snapshot to avoid early cancelation of the selector.
+
+```jsx
+const initialSnapshot = snapshot_UNSTABLE();
+initialSnapshot.retain();
+```
+
 ### Clearing all selector caches
 
 ```jsx


drarmstr:
@vntw - Yeah, probably not best to rely on the example in the Snapshot docs since that's based on the `useRecoilCallback()` hook.  Worthwhile for clarity and simplicity to have another section after [this one](https://recoiljs.org/docs/guides/testing#example-jest-unit-testing-selectors) with basically just the `retain()` line and explanation as you suggest.

vntw:
Thanks for the feedback @drarmstr, I hope the last change fits better 🙂

If I may ask, I'm not too familiar with memory management in Recoil/JS, is it necessary to release the snapshot in a jest test or will this be freed after each test anyway? The docs you linked for `useRecoilCallback` use `release()` which makes sense in a long-running app but I'm not sure here - would you recommend it?

drarmstr:
Thanks @vntw !

It's not really necessary to release for unit tests right now and most garbage collection hasn't been enabled yet.

Issue #1604, opened 2022-02-13
Debounce callback with version 0.6.0


xotahal:
Hey folks! I am not sure if this is an issue or wanted behaviour. But out of curiosity. I created this [codesandbox](https://codesandbox.io/s/xenodochial-kalam-c3g5u?file=/src/App.js:830-847) where we have one atom and two buttons. Atom has a default value set to `first`. When you press any button it will change the value of the atom and then in `useEffect` it calls `useRecoilCallback` where it gets the value from the atom and console log it.

1. Without `debounce` it sets the value to `second`. Then in `useEffect` it correctly prints out the `second` as a value.
2. The debounced button sets the value to `third`. Then in `useEffect` it prints out the previous value of the atom (either `first` or `second`). But not the `third`. As I would expect.

It works with 0.5 version.

I think it could be related to this change https://github.com/facebookexperimental/Recoil/pull/1501. Any chance someone could explain to me what exactly happened?

Thank you 🙏

xotahal:
Here is maybe a bit more "real world" [example](https://codesandbox.io/s/nice-golick-9921y?file=/src/App.js). Where I created `AmountControl` component. It has debounced `onChange`. Once it gets to the parent it changes the state.

There are two `useEffect`s, one of them uses debounced `useRecoilCallback` (that works fine) and the other one calls the `useRecoilCallback` without debounce. This one prints always previous value of the state.

drarmstr:
It's not related to #1501, but I think I have a fix.

PR #1610, opened 2022-02-15
useRecoilCallback() snapshot gets the latest state

drarmstr:
Summary:
NOTE: This is a breaking change, but the previous approach had a bug.

Previously, `useRecoilCallback()` would attempt to provide a `Snapshot` based on the state that was currently rendered.  However, there were some bugs where this was not the case.  For example, a callback called from an effect handler may or may not have had the latest rendered state based on the timing of the setter.

The new approach `useRecoilCallback()` will always try to get a snapshot with the latest state.  This should also be more intuitive for users.

Addresses #1604

Differential Revision: D34236485

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,15 @@
 # Change Log
 
 ## UPCOMING
-
-- Avoid spurious console errors from effects when calling `setSelf()` from `onSet()` handlers. (#1589)
-- Better error reporting when selectors provide inconsistent results (#1696)
-
 **_Add new changes here as they land_**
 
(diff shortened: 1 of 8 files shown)


gurkerl83:
Hi, I tried to build this pull request without success. The build mentions a circular dependency; the bundles are generated but with unresolved symbols in them. Linking recoil back to the application and running it causes an error.

drarmstr:
> @drarmstr Hi, I tried to build this pull request without success. The build mentions a circular dependency; the bundles are generated but with unresolved symbols in them. Linking recoil back to the application and running it causes an error.

Yes, it seems that our internal build avoids the circular dependency when using non-dereferencing imports while the open source build does not.  I'll have to break the circular dependency another way.

Issue #1606, opened 2022-02-14
[Feature Request] Customizable control over node to node subscriptions


FokkeZB:
> **NOTE**: This [started](https://github.com/facebookexperimental/Recoil/issues/1416#issuecomment-1037038560) as part of #1416 before [we realized](https://github.com/facebookexperimental/Recoil/issues/1416#issuecomment-1038849014) that we were talking about two different use cases.

We (at Zapier) have a need to be able to use values from multiple other selectors/atoms in a selector's `get`, while **some** of them should **not** trigger the selector to re-evaluate when their values change. AFAIK this is currently impossible.

A concrete example is that while for an API request made in a selector we need both X and Y, we only want to re-evaluate the selector and thus redo the query when X changes, or when we manually tell it to (using `useRecoilRefresher`).

I see 2 options for implementing this:

1. Add a second, `options.subscribe` argument to `get()` which, if set to `true` does not subscribe the selector to updates of the state passed to `get()`. This seems to come down to skipping [this line](https://github.com/facebookexperimental/Recoil/blob/708fcfe93a32beab35b6a9345a05dd35ab736a08/packages/recoil/recoil_values/Recoil_selector.js#L703), although it would probably be better to still track the dependency (e.g. so it can be displayed in graphs that devtools generate), but with a flag to later skip [this line](https://github.com/facebookexperimental/Recoil/blob/874328b65f03e6c3f117864355e5c88f9c50930b/packages/recoil/core/Recoil_RecoilValueInterface.js#L244) (and make a devtool e.g. render the dependency as a dashed line).

```ts
const oneState = selector({
  key: 'one',
  get: async ({ get }) => {
    // This we need and want to "subscribe" to (register as dependency), so that we re-evaluate when its value changes
    const two = get(twoState);

    // This we need, but don't want to re-evaluate for
    const three = get(threeState, { subscribe: false });   

    return fetch(url, { data: { two, three } });
  }
});
```

2. Add a `noWait` like util that uses the option [proposed](https://github.com/facebookexperimental/Recoil/issues/1416#issuecomment-1038849014) for #1416 to basically always consider the value equal (`cachePolicy_UNSTABLE.equality = () => true`).

```ts
    // This we need, but don't want to re-evaluate for
    const three = get(noSubscribe(threeState));   
```

I prefer option 1) because it's more declarative and I like how it enables us to visualize the "soft dependency" in devtools.

drarmstr:
[...]

Issue #1416, opened 2021-11-13
[Feature request] Customizable control over when a selector triggers re-render


BenjaBobs:
# Proposal

I would like to be able to specify a custom function for cache comparison on an `atom`/`selector` so I can tweak the behaviour.
Let's say we have a configurable algorithm with multiple steps, and some options that pertain to a step each.
```ts
const algorithmOptions = atom({
	key: 'algorithmOptions',
	default: {
		featureA: true,
		featureB: true,
		featureC: false,
	},
});
```
Now, only step 2 options are relevant for the step 2 of the algorithm, so we make a selector to fetch only that
```ts
const step2Options = selector({
	key: 'step2Options',
	get: ({get}) => {
		const allOptions = get(algorithmOptions);

		return {
			allOptions.featureB,
			allOptions.featureC,
		};
	},
	// overwrite cache comparison, so we don't trigger unnecessary rerenders by modifying algorithmOptions.featureA
	cacheCompare: (old, new) => old.featureB=== new.featureB && old.featureC === new.featureC,
})
```
That way, we don't trigger an unnecessary rerender when doing `setOptions({ ...options, featureA: !options.featureA })`.

# Motivation

This can currently be done using an intermediate `selector`, but it's a lot of boiler plate and makes the code look clunky.
Here's a code sandbox showing the intermediate `selector` pattern: https://codesandbox.io/s/busy-cherry-u4rx6?file=/src/App.tsx

The real life use cases for this are when you get a state object from a backend, which you want to split out into multiple parts, or if you have a form for configuration of something, and you want to save the state of the form as is.

pc-erin:
This is an urgently needed feature.
Without it any compound atom triggers a rerender even when no constituent value has changed.
If that atom is something that changes frequently, like mouse position, it has a big impact on application performance.

mondaychen:
In the example, any reason why you can't use three separate `atom`s, or an `atomFamily` as feature flags?

pc-erin:
Not who you were asking, but I have an example that involves manipulating Points (`{ x: number, y: number }`), and it wouldn't make much sense to store the x and y values in separate atoms since they will only ever be accessed together.

It also wouldn't help with the issue where selectors trigger rerenders even when there are no updates because the derived state is always a new object.

```typescript
export const viewportMousePositionState = atom<Point>({
	key: 'viewportMousePosition',
	default: ZERO_POINT
})

export const gridMousePositionState = selector<Point>({
[...]

PR #1613, opened 2022-02-15
Add 'onSetValue' method to effect params for reactive cross-state updates

### LINKED ISSUE #1611: Observable Atoms / Cross-Atom dependencies in Effects
I'd like to model some cross-atom updating logic purely within the state model. This only seems possible via React hooks, or somewhat contrived workarounds.

For example, I may have an atom that other atoms depend on, such as: 
```ts
const projectState = atom<ComplexProject({ 
  key: 'project'
  default: null,
})

const selectedFileState = atom<FileSelection>({
  key: 'selectedFile',
  default: null,
  effects: [() => {
     /* I want to subscribe to projectState changes here */
  }]
})
```

A workaround I'm considering is to use at atom effect to trigger an rxjs observable.
```ts
function observableEffect<T>(observable: Subject<void>): AtomEffect<T> {
	return ({ onSet }) => {
		onSet(() => observable.next())
	}
}


const projectStateChanged= new Subject<void>()
const projectState = atom<ComplexProject({ 
  key: 'project'
  default: null,
  effects: [observableEffect(projectStateChanged)]
})

const selectedFileState = atom<FileSelection>({
  key: 'selectedFile',
  default: null,
  effects: [({setSelf}) => {
     	const subscription = projectStateChanged.subscribe(async () => {
        const project = await getPromise(projectState)
        setSelf(project?.files[0] ?? null)
	return () => subscription.unsubscribe()
  }]
})
```

But it feels like this should be part of the core API. I could see this being implemented in a couple of different ways: 

1) Top-Level Observation API

```ts
import { observeRecoilState } from 'recoil' // usable in effects or outside of react

const selectedFileState = atom<FileSelection>({
  key: 'selectedFile',
  default: null,
  effects: [({setSelf}) => {
     	const unsubscribe = observeRecoilState(projectState, (project) => {
            setSelf(project?.files[0] ?? null)
        })
	return unsubscribe
  }]
})
```

2) observeState method in effect args:

```ts
const selectedFileState = atom<FileSelection>({
  key: 'selectedFile',
  default: null,
  effects: [({observeState, setSelf}) => {
     	const unsubscribe = observeState(projectState, (project) => {
            setSelf(project?.files[0] ?? null)
        })
	return unsubscribe
  }]
})
```

darthtrevino:
This PR adds an 'onSetValue' method to the arguments object passed to effect functions. This will allow atoms to subscribe to state changes in other atoms and reactively respond. This is useful for situations where atoms are dependent, but are not derivative from each other (where a selector would be useful).

Caveat: I'm not super familiar with Flow or the Recoil internals, so I'd be glad to shift anything around to match your teams' coding patterns

Fixes #1611

### DIFF
--- a/packages/recoil/recoil_values/Recoil_atom.js
+++ b/packages/recoil/recoil_values/Recoil_atom.js
@@ -302,6 +302,11 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         if (recoilValue.key === key) {
           throw new Error('use onSet to subscribe to mutations to self');
         }
+        if (!duringInit) {
+          throw new Error(
+            'onSetValue must be used at the top of the effect scope',
+          );
+        }
         const {release} = store.subscribeToTransactions(store => {
           const {currentTree} = store.getState();
           const currentValueLoadable = getRecoilValueAsLoadable(
--- a/packages/recoil/recoil_values/Recoil_atom.js
+++ b/packages/recoil/recoil_values/Recoil_atom.js
@@ -300,12 +300,10 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         observer: (value: S) => void,
       ): void {
         if (recoilValue.key === key) {
-          throw new Error('use onSet to subscribe to mutations to self');
+          throw err('use onSet to subscribe to mutations to self');
         }
         if (!duringInit) {
-          throw new Error(
-            'onSetValue must be used at the top of the effect scope',
-          );
+          throw err('onSetValue must be used at the top of the effect scope');
         }
         const {release} = store.subscribeToTransactions(store => {
           const {currentTree} = store.getState();
@@ -389,6 +387,9 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
 
       const onSet =
         effect => (handler: (T, T | DefaultValue, boolean) => void) => {
+          if (!duringInit) {
+            throw err('onSet must be used at the top of the effect scope');
+          }
           const {release} = store.subscribeToTransactions(currentStore => {
             // eslint-disable-next-line prefer-const
             let {currentTree, previousTree} = currentStore.getState();
--- a/packages/recoil/recoil_values/Recoil_atom.js
+++ b/packages/recoil/recoil_values/Recoil_atom.js
@@ -297,6 +297,9 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         recoilValue: RecoilValue<S>,
         observer: (value: S) => void,
       ): void {
+        if (recoilValue.key === key) {
+          throw new Error('use onSet to subscribe to mutations to self');
+        }
         const {release} = store.subscribeToTransactions(store => {
           const {currentTree} = store.getState();
           const currentValueLoadable = getRecoilValueAsLoadable(
--- a/packages/recoil/recoil_values/Recoil_atom.js
+++ b/packages/recoil/recoil_values/Recoil_atom.js
@@ -135,6 +135,7 @@ export type AtomEffect<T> = ({
   ) => void,
 
   // Accessors to read other atoms/selectors
+  observe: <S>(RecoilValue<S>, (value: S) => void) => () => void,
   getPromise: <S>(RecoilValue<S>) => Promise<S>,
   getLoadable: <S>(RecoilValue<S>) => Loadable<S>,
   getInfo_UNSTABLE: <S>(RecoilValue<S>) => RecoilValueInfo<S>,
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         return getLoadable(recoilValue).toPromise();
       }
 
+      function observe<S>(
+        recoilValue: RecoilValue<S>,
+        observer: (value: S) => void,
+      ): () => void {
+        const observed = recoilValue;
+        const {release} = store.subscribeToTransactions(store => {
+          const loadable = store
+            .getState()
+            .currentTree?.atomValues.get(recoilValue.key);
+          if (loadable) {
+            const value = loadable.getValue();
+            observer(value);
+          }
+        }, recoilValue.key);
+        return release;
+      }
+
       function getInfo_UNSTABLE<S>(
         recoilValue: RecoilValue<S>,
       ): RecoilValueInfo<S> {
@@ -408,6 +426,7 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
             setSelf: setSelf(effect),
             resetSelf: resetSelf(effect),
             onSet: onSet(effect),
+            observe,
             getPromise,
             getLoadable,
             getInfo_UNSTABLE,
--- a/packages/recoil/recoil_values/Recoil_atom.js
+++ b/packages/recoil/recoil_values/Recoil_atom.js
@@ -135,7 +135,7 @@ export type AtomEffect<T> = ({
   ) => void,
 
   // Accessors to read other atoms/selectors
-  observe: <S>(RecoilValue<S>, (value: S) => void) => () => void,
+  observe: <S>(RecoilValue<S>, (value: S) => void) => void,
   getPromise: <S>(RecoilValue<S>) => Promise<S>,
   getLoadable: <S>(RecoilValue<S>) => Loadable<S>,
   getInfo_UNSTABLE: <S>(RecoilValue<S>) => RecoilValueInfo<S>,
@@ -296,18 +296,17 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
       function observe<S>(
         recoilValue: RecoilValue<S>,
         observer: (value: S) => void,
-      ): () => void {
-        const observed = recoilValue;
+      ): void {
         const {release} = store.subscribeToTransactions(store => {
-          const loadable = store
-            .getState()
-            .currentTree?.atomValues.get(recoilValue.key);
-          if (loadable) {
-            const value = loadable.getValue();
-            observer(value);
-          }
+          const loadable = getRecoilValueAsLoadable(store, recoilValue);
+          const value = loadable.getValue();
+          observer(value);
         }, recoilValue.key);
-        return release;
+
+        cleanupEffectsByStore.set(store, [
+          ...(cleanupEffectsByStore.get(store) ?? []),
+          release,
+        ]);
       }
 
       function getInfo_UNSTABLE<S>(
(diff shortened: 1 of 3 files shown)


packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
...
+        recoilValue: RecoilValue<S>,
+        observer: (value: S) => void,
+      ): () => void {
+        const observed = recoilValue;
+        const {release} = store.subscribeToTransactions(store => {
+          const loadable = store
+            .getState()
+            .currentTree?.atomValues.get(recoilValue.key);
+          if (loadable) {
+            const value = loadable.getValue();
```
drarmstr, line 306: `getValue()` could throw in error or pending states.  Check that the state is `hasValue` to avoid uncaught exceptions.

packages/recoil/recoil_values/__tests__/Recoil_atom-test.js
```diff
@@ -911,6 +911,39 @@ describe('Effects', () => {
...
+      </>,
+    );
+
+    expect(c.textContent).toEqual('10099');
+    act(() => setA(100));
+    expect(c.textContent).toEqual('100101');
+    act(() => setA(101));
+    expect(c.textContent).toEqual('101102');
+  });
+
```
drarmstr, line 946: Add test cases for resetting and subscribing to selectors.  test when selectors are in error or pending states.

Test this and `onSet()` can only be called directly from the effect function itself, not nested in another callback.

There's an "unsubscribe" unit test that should also test the handler is always unsubscribed same as `onSet()`.

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         return getLoadable(recoilValue).toPromise();
       }
 
+      function observe<S>(
+        recoilValue: RecoilValue<S>,
+        observer: (value: S) => void,
```
drarmstr, line 298: Could potentially also get the previous value and `isReset` consistent with `onSet()`.  But, that could also be added later, if needed.

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         return getLoadable(recoilValue).toPromise();
       }
 
+      function observe<S>(
+        recoilValue: RecoilValue<S>,
```
drarmstr, line 297: Needs to handle the case when `recoilValue` is actually ourselves.  Either this could redirect to `onSet()` in that case or `onSet()` could be renamed/generalized to handle any node and then a `onSet()` is just a wrapper calling that.

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
         return getLoadable(recoilValue).toPromise();
       }
 
+      function observe<S>(
```
drarmstr, line 296: Naming is always hard..  I suppose might have made sense for this to be `onSet()`, and that could have been `onSetSelf()` or something; but too late to make that change now.  Hmm, maybe this could be `onSetValue()`...

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
...
+        recoilValue: RecoilValue<S>,
+        observer: (value: S) => void,
+      ): () => void {
```
drarmstr, line 299: Should probably detect and error that this (and `onSet()` as well) can only be called directly from the effect (e.g. `duringInit`).  Otherwise it could encourage trying to setup dynamic subscriptions and things get more complicated...

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -135,6 +135,7 @@ export type AtomEffect<T> = ({
   ) => void,
 
   // Accessors to read other atoms/selectors
+  observe: <S>(RecoilValue<S>, (value: S) => void) => () => void,
```
drarmstr, line 138: Type definitions and test would also have to be added in `/typescript`

drarmstr (changes requested):
Thanks for the contribution!

We were against adding the ability for atom effects to write to other atoms due to the risk of complexity.  This looks like an incremental addition to the existing ability to poll other state, however.  Let me discuss this extension with the team.

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
...
+        observer: (value: S) => void,
+      ): () => void {
+        const observed = recoilValue;
+        const {release} = store.subscribeToTransactions(store => {
+          const loadable = store
+            .getState()
+            .currentTree?.atomValues.get(recoilValue.key);
+          if (loadable) {
+            const value = loadable.getValue();
+            observer(value);
```
drarmstr, line 307: When at atom is set to an equivalent (via reference equality) value then it short-circuits the set operation and subscribers such as `onSet()` are not called.  This introduces the ability to subscribe to selectors as well as atoms.    Currently selectors will also only cause subscribing components to re-render if they are changed to a non-reference-equality value and there is a lot of attention paid to this by users.  (Note there are also requests for more fine-grained control over equality (#1416) and skipping downstream node re-evaluation (#1606) as well).  Thought should also be made if something equivalent can be done here.

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
...
 
+      function observe<S>(
+        recoilValue: RecoilValue<S>,
+        observer: (value: S) => void,
+      ): () => void {
+        const observed = recoilValue;
+        const {release} = store.subscribeToTransactions(store => {
+          const loadable = store
+            .getState()
+            .currentTree?.atomValues.get(recoilValue.key);
```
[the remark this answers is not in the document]
darthtrevino: I pushed an update to this inner loop that looks a bit cleaner. One thing we'll want to consider is the case where the loadable has an error - should we have a second callback for errors in the `observe` signature?

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -292,6 +293,23 @@ function baseAtom<T>(options: BaseAtomOptions<T>): RecoilState<T> {
...
+          if (loadable) {
+            const value = loadable.getValue();
+            observer(value);
+          }
+        }, recoilValue.key);
+        return release;
```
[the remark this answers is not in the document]
darthtrevino: That makes sense, done

packages/recoil/recoil_values/Recoil_atom.js
darthtrevino: Yep, makes sense

packages/recoil/recoil_values/Recoil_atom.js
darthtrevino: Yep, as pointed out above I think we'll need access to the defaultLoadable on the upstream RecoilValue, which may require some refactoring.

packages/recoil/recoil_values/Recoil_atom.js
darthtrevino: Ahh, right - I updated this to check for key equality and throw an error if if's a self-observe.

packages/recoil/recoil_values/Recoil_atom.js
darthtrevino: I'm open to changing this, that suggestion seems to at least rhyme with the current naming scheme. I'll push an update with your suggestion.

packages/recoil/recoil_values/Recoil_atom.js
darthtrevino: Aye aye

packages/recoil/recoil_values/Recoil_atom.js
darthtrevino: I added an error condition in onSetValue- I'd be glad to add one in onSet if you prefer, but I don't want to blow the scope of the PR up too much

packages/recoil/recoil_values/__tests__/Recoil_atom-test.js
darthtrevino: @drarmstr  - I ran into an issue with `store.subscribeToTransactions` when passing in the key of a selector. In that case it does not emit events and the observing state doesn't update. I can work around this for selectors by subscribing to the global event stream, but I suspect that will be chattier than what we want, and it breaks a test case regarding resetting an atom that uses onSetValue

darthtrevino:
Moving this comment to the top-level thread:

@drarmstr - I ran into an issue with store.subscribeToTransactions. When passing in the key of a selector. In that case it does not emit events and the observing state doesn't update. I can work around this for selectors by subscribing to the global event stream, but that is chattier than what we want, and it breaks the case when you have an atom that subscribes to another value, and you reset its state.

I put some notes in the test file and the `onSetValue` function. I suspect that this is a bug with the store.

drarmstr:
> Moving this comment to the top-level thread:
> 
> @drarmstr - I ran into an issue with store.subscribeToTransactions. When passing in the key of a selector. In that case it does not emit events and the observing state doesn't update. I can work around this for selectors by subscribing to the global event stream, but that is chattier than what we want, and it breaks the case when you have an atom that subscribes to another value, and you reset its state.
> 
> I put some notes in the test file and the `onSetValue` function. I suspect that this is a bug with the store.

Ah, yes, that seems like a big concern.  We wouldn't want to subscribe to global state changes, as that would call too often and would also be a scalability performance concern.  Actually, the `dirtyAtoms` in the store only records atoms that have changed, not selectors, as they are derived.  So, it's not really a bug in the store, but rather recording dirty selectors would be a new feature.  What might work in `sendEndOfBatchNotifications()` would be using `getDownstreamNodes()` to get selectors that are potentially updated.  However, I think with profiling we've seen that `getDownstreamNodes()` is already a performance scalability concern (See `Recoil_perf-test.js`) and using it there for each atom for each batch would be problematic.  Options around this might be recording dirty selectors along with dirty atoms and/or optimization of `getDownstreamNodes()`.  Issues from #1416 and #1606 are also relevant here.

packages/recoil/recoil_values/Recoil_atom.js
```diff
@@ -134,6 +134,9 @@ export type AtomEffect<T> = ({
     (newValue: T, oldValue: T | DefaultValue, isReset: boolean) => void,
   ) => void,
 
+  // Subscribe to changes in other recoil values
+  onSetValue: <S>(RecoilValue<S>, (value: S) => void) => void,
```
CollinMonahanLab49, line 138: 👍

PR #1651, opened 2022-03-04
Optimize Selector Tree Cache

drarmstr:
Summary: Optimize the selector tree cache to use an iterative vs a recursive algorithm.  This hurts readability, but is important to allow more selector dependencies than the JavaScript callstack depth limit.

Differential Revision: D34633750

### DIFF
--- a/packages/recoil/core/__tests__/Recoil_perf-test.js
+++ b/packages/recoil/core/__tests__/Recoil_perf-test.js
@@ -13,6 +13,7 @@
 import type {Loadable, RecoilState, RecoilValue} from '../../Recoil_index';
 
 const {atom, selector} = require('../../Recoil_index');
+const {waitForAll} = require('../../recoil_values/Recoil_WaitFor');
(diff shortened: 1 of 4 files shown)


packages/recoil/core/__tests__/Recoil_perf-test.js
```diff
@@ -26,11 +27,15 @@ const ITERATIONS = [1]; // Avoid iterating for automated testing
 // const ITERATIONS = [10, 100, 1000, 10000];
 // const ITERATIONS = [10, 100, 1000, 10000, 100000];
 
-function testPerf(name: string, fn: number => void) {
+function testPerf(
```
thomaszdxsn, line 30: how does this test case assure performance is good?

drarmstr: It doesn't for automated testing.  But, it's there for at least being able to manually run performance tests.

PR #1653, opened 2022-03-06
[docs/korean] add translate useRecoilRefresher.md to Korean

hellojaehyeok:
Currently, there is no "useRecoilRefresher.md" file translated into Korean, so it was produced.

If you access "https://recoiljs.org/ko/docs/api-reference/core/useRecoilRefresher/", it's the English version.


drarmstr:
Thanks @hellojaehyeok!

Issue #794, opened 2020-12-14
Setting an object value inside an atom returns "Cannot assign to read only property 'xxxxx' of object '#<Object>'"


pixel-mattp:
I am trying to edit a key of 'activeCourse' inside an object of an atom value.  However, whenever i do so it returns the error: 

`Cannot assign to read only property 'activeCourse' of object '#<Object>'`

The atom is called 'trainingAtom' and default value is an array of objects:

```
[
  {
     _id: '5fc42caa244ccadea8e2f263',
     activeCourse: 'r4r3545243j',
     activeDetail: null,
   },
   {
      _id: '5fc42caa244ceecadea8e2f263',
      activeCourse: null,
      activeDetail: null,
    },
]
```

I am using the following function within my component:

```
const [test, setTest] = useRecoilState(trainingAtom);

useEffect(() => {
        let _test = [...test];
        const index = _test.findIndex((val) => val._id === id);
        _test[index].activeCourse = 123456;
        setTest(_test);
}, [id]);
```

This function works if I use normal React state - but anything inside Recoil throws the error.  
Note, I have also tried to use the `set` function inside a selector but it gives the same error.  

What am I doing wrong?

bentron2000:
I have also encountered this issue.

cbuv:
have you tried using updater instead of value? 
like this: 
```
export const Test = ({id}: any) => {
  const [test, setTest] = useRecoilState(trainingAtom);

  useEffect(() => {
    setTest((old: any) => {
      let _test = [...old];
      const index = _test.findIndex((val) => val._id === id);
      //@ts-ignore
      _test[index].activeCourse = 123456;
      return _test;
    });
  }, [id, setTest]);
  return (<div>{JSON.stringify(test)}</div>);
}
```

pixel-mattp:
@cbuv - this works.  Seems strange that it doesn't seem to like the spread operator from state directly.  Thanks for this.

cbuv:
Yeah, the docs could be more clear about this. 

From the docs: (useRecoilState)
This API is similar to the React useState() hook except it takes a Recoil state instead of a default value as an argument. It returns a tuple of the current value of the state and a setter function.

I think the latter part should include how it differs from React, even if setter is best practice. Primitives can use value, but not arrays, guess setter must be used with array of primitives too

drarmstr:
Clarified in docs #800 that atom/selector values are frozen.  

[...]

PR #1679, opened 2022-03-17
Add readonly TS types for atoms and selectors to prevent mutability bugs

csantos42:
Because we perform a deep freeze to recoil values, mutating those values at runtime (during dev) causes our code to throw, which can lead to confusing errors. In my case, I was calling `arr.sort()` and `arr.reverse()` in a component, and because that mutates the underlying array my app was crashing in dev. 

After this change the result of reading atom/sel values from hooks will not allow for mutation of the returned values. This conflicts when `dangerouslyAllowMutability` is enabled, but that can be fixed in userland by having a wrapper over the hooks to expose mutable types (I think the TS types should match the default behavior of disallowing mutations by default).

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -116,9 +116,16 @@
  }
  export type AtomOptions<T> = AtomOptionsWithoutDefault<T> | AtomOptionsWithDefault<T>;
 
- type DeepReadonly<T> = {
-  readonly [P in keyof T]: DeepReadonly<T[P]>
- };
+ /**
+  * Will work for all types, but do not add a special case for Arrays specifically as
+  * that will be handled in the last case (in the same condition that handles objects).
+  * If we add a special case for arrays we will lose tuple types.
+  */
+ type DeepReadonly<T> =
+   T extends ((...args: any[]) => any) | boolean | number | string | null | undefined ? T :
+   T extends Map<infer K, infer V> ? ReadonlyMap<DeepReadonly<K>, DeepReadonly<V>> :
+   T extends Set<infer S> ? ReadonlySet<DeepReadonly<S>> :
+   {readonly [P in keyof T]: DeepReadonly<T[P]>};
 
  /**
   * Creates an atom, which represents a piece of writeable state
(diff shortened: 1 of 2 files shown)


typescript/index.d.ts
```diff
@@ -116,10 +116,14 @@
  }
  export type AtomOptions<T> = AtomOptionsWithoutDefault<T> | AtomOptionsWithDefault<T>;
 
+ type DeepReadonly<T> = {
+  readonly [P in keyof T]: DeepReadonly<T[P]>
```
mondaychen, line 120: Does this type accept arrays?

drarmstr: Are you aware of any Flow equivalent?

csantos42: @mondaychen it does! Although I was missing a few special cases (`Map`,`Set`, functions) so I pushed an updated version. And @drarmstr I think in theory it would be possible but potentially painful since flow makes it hard to do conditional types, something like this should work (but seems to be an old gist, maybe there's a better way now): https://gist.github.com/karlhorky/99ecb289504dd96d1510ee3961aeeac0  , and it doesn't seem to have a special case for `Map` and `Set`

klcsantos: Ah good question, in terms of custom classes it'll treat the properties as frozen, but methods that mutate the object are still allowed at the type level since we can't know ahead of time which methods in a custom class mutate the object. I think that might be an ok tradeoff though

csantos42: @drarmstr Ah good question, in terms of custom classes it'll treat the properties as frozen, but methods that mutate the object are still allowed at the type level since we can't know ahead of time which methods in a custom class mutate the object. I think that might be an ok tradeoff though

drarmstr: @csantos42  Yeah, it's not perfect but seems like it could help.  The Flow team didn't have suggestions for a recommended helper for this.  I guess the primary concern is how much this might impact `dangerouslyAllowMutability` usage?

csantos42: good point @drarmstr , the experience for mutability should be taken into account so I overloaded the constructors to only make the value readonly when `dangerouslyAllowMutability` is false/not set. (and added corresponding unit tests)

drarmstr: Nice!

drarmstr:
This would help avoid user confusion around #794

csantos42:
@mondaychen I think this was never merged but it's still a relevant change as far as I'm aware (values read from state should be treated immutably)

PR #1703, opened 2022-03-29
Better typing for maps, sets, and classes to be used in family parameters

drarmstr:
Summary:
Fix the typing to explicitly allow `Map<>` and `Set<>`.

Fix the typing for `atomFamily()` and `selectorFamily()` parameters to better allow for objects that are class instances that have a `toJSON()` method to be used, as they are serializable.

Previously it required `toJSON()` to return a `string`, but it could really return any serializable primitive or object.

Previously it used the arrow syntax for the `toJSON()` method which requires that the method is bound to the instance for dereferencing `this` during execution.  That should not be a requirement, though, as we always call it relative to the instance itself.  This change allows custom user classes to be used.  For example:

```
class MyClass {
  _name: string;

  constructor(name: string) {
    this._name = name;
  }

  toJSON(): string {
    return this._name;
  }
}
```

There is no functional change with this diff, only type fixes.

Differential Revision: D35233518

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## UPCOMING
 **_Add new changes here as they land_**
 
+- Update typing for family parameters to better support Map, Set, and classes with `toJSON()`. (#1709, #1703)
+
 ## 0.7 (2022-03-31)
 
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## UPCOMING
 **_Add new changes here as they land_**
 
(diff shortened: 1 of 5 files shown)


victoriafrench:
classes are expressly forbidden in atoms as they are transformed into an immutable object. this is a constraint imposed by facebook. it is also forbidden for recoil to support class components as they are being prepared to be dropped from react. everything runs under reacts future of pure functional paradigm which forbids mutation, which is what classes do.

victoriafrench (commented):
none of the examples here are valid uses of a selectorFamily. selectorFamilies monitor atoms.

we should discuss on how these work and when they used.

PR #1753, opened 2022-04-25
Fix nodenext module support from TypeScript 4.7

wight554:
* as per https://github.com/microsoft/TypeScript/issues/48235#issuecomment-1067363731


wight554:
As temporal local solution be shimmed with:
```
declare module 'recoil' {
  export * from 'node_modules/recoil/index.js';
}
```

PR #1769, opened 2022-05-06
Prepare build scripts for releasing multiple packages

mondaychen:
As titled.

The new build scripts now build both recoil and recoil-sync to `build/` folder. We will release from there.
To test, run `yarn build-all`, then check `build` folder.
You should be able to release from `build/recoil` and `build/recoil-sync`

Next steps (after this is merged):
- write a README.md for recoil-sync
- Then we can test & publish the first version of recoil-sync! 😉
- add `transit-js` to `dependencies` and update lock file
- update CI
- remove old build scripts & update release process wiki


packages/recoil-sync/package-for-release.json
```diff
@@ -0,0 +1,20 @@
...
+  "main": "cjs/index.js",
+  "module": "es/index.js",
+  "react-native": "native/index.js",
+  "unpkg": "umd/index.js",
+  "types": "index.d.ts",
+  "files": ["umd", "es", "cjs", "native", "index.d.ts"],
+  "repository": "https://github.com/facebookexperimental/Recoil.git",
+  "license": "MIT",
+  "dependencies": {
+    "transit-js": "^0.8.874"
```
mondaychen, line 15: Is there something I missed?

Issue #289, opened 2020-06-09
There is a way to update states on recoilJS outside of component?


orhalimi:
So I'm trying recoilJS for a js game that I am building and it pretty neat, but the need to update atoms from components only feels like a limitation.

To create a game loop, I put all the logic on empty component so I will be able to read and write states. Even if I will construct the logic outside of the component, I will need especially move different stats around all the time. There is a way to update atoms outside of react component (not via hooks)?

pankerit:
it would be great,
and to subscribe outside of component would be nice

ehahn9:
related: what would the idiom be for creating async selectors which also subscribe to an external data source and trigger component renders when the (external) data changes?

- contrived: a selector which reports the window dimensions and listens via `window.onresize` etc.

- not too contrived (my real-world case): a selector which returns a field from a firestore document, the contents of which might change externally (notified by `onSnapshotChanged`). This second case is really cool for recoil because the fetched document would be shared state (e.g. key === document.id) so you'd be pretty sure not to unneccessarily re-fetch data...

I'm sure I'm just dim, but it wasn't clear how you'd construct the selector getter in those cases?

ehahn9:
@Shmew that sounds like an excellent direction. Thanks for the pointer. In this approach, any thoughts on how the hook would  know when to unsubscribe from it's listener? I didn't see a way for `useRecoilCallback` to provide a cleanup to its effect so I'd worry that the hook would keep listening for changes and updating the atom long after all the consumers were unmounted? Apologies if I'm not following fully.

Shmew:
Yeah, you would need to implement some sort of mechanism to cancel the subscriber. Probably via a clean-up `useEffect` in the component.

n8sabes:
@davidmccabe, I'm finding numerous issues requesting access to atoms and selectors outside of components, but issues on the matter get closed without a clear path forward / example.

Like @atulmy [authentication inquiry](https://github.com/facebookexperimental/Recoil/issues/73#issuecomment-631391642), I cannot figure out how to migrate our authentication logic from Redux to Recoil. We need to call authentication helper functions in our auth library from various components whereby the Recoil state gets updated upon network call completion in the auth library. After a few days of work migrating from Redux to Recoil, I've completely stalled. I am invested and really want Recoil to work.

If the answer is, "Recoil is not an end-to-end state manager", that's fine but need to get clarity on where to and where not to use Recoil.

Thanks for any thoughts / examples on this.

[...]

PR #1826, opened 2022-05-31
[Proposal] Include Recoil Nexus in the package.

luisanton-io:
As requested by multiple users<sup>[[1]](https://stackoverflow.com/a/69361551/11783958)</sup><sup>[[2]](https://github.com/facebookexperimental/Recoil/issues/289#issuecomment-901624892)</sup><sup>[[3]](https://github.com/facebookexperimental/Recoil/discussions/1816)</sup> and considering the recent traction of [Recoil Nexus](https://www.npmjs.com/package/recoil-nexus), hopefully we can include this feature directly in the package.

I am not particularly happy I had to extract `RecoilRoot` from its original module to avoid a circular dependency with `useRecoilCallback`.
This has led me to rename the exports everywhere from `/Recoil_RecoilRoot` to `/Recoil_RecoilNexus`, tests included.

If this is a big deal any hint to avoid doing so is welcome.
Cheers,
L

### DIFF
--- a/packages/recoil/core/Recoil_RecoilNexus.js
+++ b/packages/recoil/core/Recoil_RecoilNexus.js
@@ -89,7 +89,7 @@ type Props =
     };
 
 function RecoilRoot(props: Props): React.Node {
-  const {override, children, ...propsExceptOverride} = props;
+  const {override, children, ...propsExceptOverrideAndChildren} = props;
 
   const ancestorStoreRef = useStoreRef();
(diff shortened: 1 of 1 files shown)


drarmstr:
Thanks for this contribution @luisanton-io!  Though, we should explore some of the nuances of the interface for consideration.  It seems like this API would have race conditions in case there are multiple `<RecoilRoot>`'s used..

luisanton-io:
I see. Hopefully I will have some time to check into this soon. Thanks for the heads up

PR #1858, opened 2022-06-09
Prepare release of refine as its own package

mondaychen:
As titled...

Alpha version released as
- refine: https://www.npmjs.com/package/@recoiljs/refine/v/0.1.1-alpha.1
- recoil-sync: https://www.npmjs.com/package/recoil-sync/v/0.1.1-alpha.1

Will need to update docs later

### DIFF
--- a/packages/recoil-sync/package-for-release.json
+++ b/packages/recoil-sync/package-for-release.json
@@ -1,6 +1,6 @@
 {
   "name": "recoil-sync",
-  "version": "0.1.0",
+  "version": "0.1.1-alpha.1",
   "description": "recoil-sync provides an add-on library to help synchronize Recoil state with external systems",
   "main": "cjs/index.js",
   "module": "es/index.js",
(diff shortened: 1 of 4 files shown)


packages/recoil-sync/package-for-release.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "recoil-sync",
-  "version": "0.1.0",
+  "version": "0.1.1-alpha.1",
```
drarmstr, line 3: There hasn't been a `0.1.0` release yet, so I think you could still use `0.1.0-alpha.1`

packages/refine/package-for-release.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@recoiljs/refine",
+  "publishConfig": {
```
drarmstr, line 3: I don't think the other `package.json`'s had this.  Is it necessary or should we add it to the others?  Or is it needed here because it is a scoped package?

mondaychen: It's because it's a scoped package. Scoped packages are private by default

packages/recoil-sync/package-for-release.json
mondaychen: Ah you are right. I'll change it to 0.1.0. When we officially release it we can call it 0.1.0, but I won't publish a 0.1.0-alpha.1 unless it become necessary

PR #1931, opened 2022-08-03
Refine - remove 'native' from package.json

jake-nz:
There is no `native` folder in the release but it was referred to in the package file, which causes errors for react-native.


jake-nz:
> @drarmstr has imported this pull request.

What has happened here? Does that mean it's being accepted by some process outside Github?

drarmstr:
Yes, it was imported to our internal source repository which is the source of truth for the code base.  When changes land in that repository then they are replicated back out to the GitHub repository.

Issue #733, opened 2020-11-11
[SSR][NextJS] Duplicate atom key during development and during production build in nextjs


janus-reith:
Both during local development and also when building a production build, I’m getting a duplicate atom key for each key and each page using it after the first one.

I put together a quick sandbox to demonstrate the issue:
https://codesandbox.io/s/flamboyant-sea-tqlky

The errors show up in the build log, a quick way to test is using the integrated vercel deployment button.
Looking at the default nextjs example, I can‘t spot any special settings there that would prevent whatever duplication is going on from happening: https://github.com/vercel/next.js/tree/canary/examples/with-recoil
That example however only makes use of recoil state on one page.

alexilyaev:
Also mentioned in:  
https://github.com/facebookexperimental/Recoil/issues/213#issuecomment-659349900

This is how it looks like in the Vercel deployment build:
![image](https://user-images.githubusercontent.com/480469/98931775-231bc580-24e7-11eb-956f-a4c57cc7b61f.png)

leonardomgt:
Has this been solved? I'm having the same issue.

janus-reith:
@leonardomgt I was not able to solve this yet.

@drarmstr Tagging you as you seem to be a main contributor to this project, any chance you could have look at this? Thanks!

drarmstr:
@janus-reith - Sorry, I'm not very familiar with Next.JS or SSR.

alexilyaev:
> @janus-reith - Sorry, I'm not very familiar with Next.JS or SSR.

Next.js has a concept of Pages. Imagine an SPA with multiple entry points, like an Admin and a Client App, both importing a module that declares a Recoil `atom`. So it doesn't have to be Next.js related.  
Essentially it's a single Node.js process that executes JS files (build process) that eventually declare the same `atom` several times.

In development, when a file is changed, Next.js re-builds the relevant page entry file.  
Because it's the same Node.js process, the `atom` has already been declared.  
The same thing can happen with HMR when the file change triggers a rebuild of the whole file, or even when the `atom` is declared inside a component lifecycle/hook and only that is being hot-replaced.

Basically, I can't think of a solution to this problem besides providing a configuration setting to disable this check/warning.  
At a minimum, developers should be able to disable it.  
In Next.js or SSR that would be when running on the server.  
In HMR, that would be when not in `production` env.

A zero config alternative might be tempting. E.g. relying on `NODE_ENV=production` like React does for it's optimizations.  
[...]

Issue #1912, opened 2022-07-19
Disable "Recoil: Spent N ms computing a cache key?


FezVrasta:
Is it possible to disable these warnings? On CI we fail the build on console messages and this makes the builds flaky.

mondaychen:
This is a dev mode only warning. Is it possible to build in prod mode?

FezVrasta:
We run the tests and I guess they use the dev flag by default. I'm not sure if it's possible to change that without breaking anything around Jest and such

mondaychen:
Why do you need to fail the build on console messages?
If you have to fail on console, I still don't think it's a good idea to disable console output in recoil. You are likely going to to encounter this issue again with other libs
If you use babel, try this plugin? https://www.npmjs.com/package/babel-plugin-transform-remove-console

FezVrasta:
Sorry for the delay in my reply.

We disallow console.logs so that when we run tests we ensure all the console.error, console.warn, etc... are properly handled.

For example, if you do expect your code to throw a console.warn we ask the developers to do `expect(console.warn).toHaveBeenCalledWith('message')` so that we are sure the behavior is consistent.

The problem with this Recoil warning is that it's performance sensitive, so if the runner is more busy than usual it will trigger it, otherwise it will not. As this is not an issue in real-life scenarios we would like to disable it, the problem is that the warning doesn't have a known prefix or code we can use to "whitelist" it, so it's very hard to find a way to get around it.

FezVrasta:
?

PR #1974, opened 2022-08-19
fix: temporary implementation fixes for `isFastRefreshEnabled

AkifumiSato:
A duplicate atom key warning is output during development(ref: #733).
The implementation that determines hot modules is tentative and always returns `false`, so the output is always present during development.
I propose to set it `true` temporarily.


drarmstr:
This would dangerously suppress the warning from catching real bugs.

koichik:
@drarmstr Of course I understand the importance of the warning. However, many React developers use Next.js, RCA, etc. and they enable Fast Refresh in the dev mode. So we can usually only ignore the warning.
Would you please consider the isFastRefreshEnabled returns true by default until you can correctly detect reloads due to Fast Refresh?

drarmstr:
A complete disabling by default is too dangerous overall.  Perhaps a mechanism to be able to explicitly suppress the warning push might be considered.  @mondaychen also suggested a Babel plugin might provide this functionality in #1912

Issue #1900, opened 2022-07-02
<RecoilURLSync> doesn't reset state if param is not present


koichik:
`RecoilURLSync{JSON,Transit}` can take a `param` if the `location`'s `part` is `'queryParams'`.
However, if the new URL doesn't have the part, `RecoilURLSync{JSON,Transit}` doesn't reset a state that syncing it.

for example,

Atom:
```
const nameState = atom<string>({
  key: 'name',
  default: '',
  effects: [syncEffect({ refine: string() })],
})
```

App:
```
function MyApp({ Component, pageProps }: AppProps) {
  return (
    <RecoilRoot>
      <RecoilURLSyncJSON
        location={{ part: 'queryParams', param: 'state' }}
      >
        <Component {...pageProps} />
      </RecoilURLSyncJSON>
    </RecoilRoot>
  )
}
```

1. access to:
`http://localhost:3000/users/1`

2. change name state to `'Foo'`, then URL is changed to:
`http://localhost:3000/users/1?state=%7B%22name%22%3A%22Foo%22%7D`

3. navigate to:
`http://localhost:3000/users/2`

expected: the `name` state is reset to the default (empty string).
actual: the `name` state still `'Foo'`.

I think that this line should return empty `Map()` instead of `null` when `stateStr` is `null`

https://github.com/facebookexperimental/Recoil/blob/1a47a17a577bf9fc96479a1f700368f4817a166d/packages/recoil-sync/RecoilSync_URL.js#L76

koichik:
The example on Next.js is here.
https://github.com/recruit-tech/recoil-sync-next

## Install

```bash
git clone https://github.com/recruit-tech/recoil-sync-next
cd examples/url-sync-json/
yarn
yarn dev
```

# location={{ part: 'queryParams' }} (no problem)

`examples/url-sync-json/pages/_app.tsx`:

      <RecoilURLSyncJSONNext
        storeKey="url-json-store"
        location={{ part: 'queryParams' }}
      >

open http://localhost:3000/

![wo-params-1](https://user-images.githubusercontent.com/355825/179668926-d9db21a3-61b6-43c6-9923-366366edc7ad.png)

1. modify the form.
2. the URL will be updated.
![wo-params-2](https://user-images.githubusercontent.com/355825/179668945-c635199a-801c-4b94-9a02-a0ca1cfde6bd.png)

3. click the "Top" link.
[...]

Issue #1975, opened 2022-08-19
`urlSyncEffect` not syncing after an external url change


nichita-pasecinic:
If using `urlSyncEffect` effect I assume that any mutation to recoil atom would change the url (In may case just a query param key), but as well in case of a navigation from client side with a link, or `useNavigate` would update the atom value

```jsx
const navigate = useNavigate();
navigate('/something'); // <-- removes the current query params (linked with recoil atom), but does not reset them 
```

nichita-pasecinic:
Are there any ways I could remove the queryParam from the url bar and have the atom being reset to the default state ?

drarmstr:
Recoil Sync attempts to sync bidirectionally, so changing an atom will update the URL and changing the URL will update the atom.  (Note #1900 which may also be related to your issue, though that is fixable)  It relies on subscribing to the browser `popstate` event for URL changes.  This covered the usecase of using the Browser "back" button which was our priority.  However, it appears this event is not triggered with all forms of navigation.  There are also a `hashchange` event, but I think someone tried testing with that without much luck.  I'm going to assume by `useNavigate()` you are referencing React Router.  I don't know their internals, but assume they are using `history.replace();` or `history.pushstate();`.   As a workaround some have either manually triggered the `popstate` event or make extra `history.pushstate(); history.popstate();` calls.  Another solution might be to update the URL sync logic with polling, if someone wanted to add that.

salvoravida:
@drarmstr https://github.com/remix-run/history/blob/dev/packages/history/index.ts

Maybe recoil URL sync could use well-tested history wrapper: npm "history" package (from the team of react-router)

drarmstr:
@salvoravida Looking at the source of `history` package it looks like they also just rely on the `popstate` and `hashchange` events.  I can add the `hashchange` event for more coverage.  Though, I did actually manually test out some external URL changes and found even with just `popstate` it was working with relative links with `href` that changed either the search params or anchor hash and a button that changed `location.hash` with Chrome at least.  Not sure what mechanism `useNavigate()` uses.

https://codesandbox.io/s/heuristic-elion-suen5c?file=/src/App.tsx

PR #1976, opened 2022-08-19
fix: Reset atoms when param is removed from URL when `queryParams` with `param` is specified

AkifumiSato:
## sumary
resolve: #1900 
If `queryParams` and `param` are specified and there are no parameters in the URL, the update process may not be performed.
The reason is that null is assigned [here](https://github.com/facebookexperimental/Recoil/blob/0.7.5/packages/recoil-sync/RecoilSync_URL.js#L75), and [the branch](https://github.com/facebookexperimental/Recoil/blob/0.7.5/packages/recoil-sync/RecoilSync_URL.js#L248-L250) is not passed.

## note
I was going to write a test, but the current test relies on [RecoilSync_MockURLSerialization.js](https://github.com/facebookexperimental/Recoil/blob/0.7.5/packages/recoil-sync/__test_utils__/RecoilSync_MockURLSerialization.js#L92), so it is difficult to write a reproducible test for this bug.

### DIFF
--- a/packages/recoil-sync/RecoilSync_URL.js
+++ b/packages/recoil-sync/RecoilSync_URL.js
@@ -72,7 +72,7 @@ function parseURL(
       const {param} = loc;
       if (param != null) {
         const stateStr = searchParams.get(param);
-        return stateStr != null ? wrapState(deserialize(stateStr)) : null;
+        return stateStr != null ? wrapState(deserialize(stateStr)) : new Map();
       }
       return new Map(
         Array.from(searchParams.entries()).map(([key, value]) => {


drarmstr:
@AkifumiSato - That mock URL serialization doesn't need to necessarily be used by all tests or can be extended.  There is also a browser interface abstraction that may be useful for testing as well.

AkifumiSato:
@drarmstr Yes, I tried to write a test without mock, but it did not work.
This is the diff when I tried it on another branch.
https://github.com/AkifumiSato/Recoil/commit/a58eb75858a1afc0f012ee457734b6846984dac4#diff-74f521e2004fc6640581dbfffa78e3cf816c4b8b55c81ab8f58d21873af6981cR193

This is what I found when I tried this test case.
- `pushState` after rendering in a test case does not change the rendering(fail L.197).
- `await flushPromisesAndTimers` causes URL to revert back to before push (fail L.196).

For these reasons, I gave up implementing this tests.
Maybe I am not doing it right, but I could not figure out the cause.
Would you advise?

drarmstr:
I think that's just because actions like `history.pushState()` don't trigger a `popstate` browser event.  (See #1975). I've added the following updated test to the PR (No need for you to update the PR)

```
test('Remove parameter', async () => {
  const loc = {part: 'queryParams', param: 'param'};

  const atomA = atom({
    key: 'recoil-url-sync remove param',
    default: 'DEFAULT',
    effects: [syncEffect({itemKey: 'item', refine: string()})],
  });

  const container = renderElements(
    <TestURLSync location={loc}>
      <ReadsAtom atom={atomA} />
    </TestURLSync>,
  );
  expect(container.textContent).toBe('"DEFAULT"');

  history.replaceState(null, '', encodeURL([[loc, {item: 'SET'}]]));
  history.pushState(null, 'void');
  history.back();
  await flushPromisesAndTimers();
  expect(container.textContent).toBe('"SET"');

  // clear all query params from the URL to confirm it resets the atoms.
  history.replaceState(null, '', location.origin);
  history.pushState(null, 'void');
  history.back();
  await flushPromisesAndTimers();
  expect(container.textContent).toBe('"DEFAULT"');
});

```

mondaychen:
This is strange. Why does this PR has no CI results?

drarmstr:
> This is strange. Why does this PR has no CI results?

Because it was an external submission which I then extended with tests and test fixes in www, but because it originated with the external PR I can't export those edits back out to the PR again to run the CI.  Seems like a limitation in our syncing mechanism.

PR #2033, opened 2022-09-23
[docs]  make RecoilURLSyncTransit 'handlers' a stable constant, …

wd-fb:
…unstable value will show warning in the future version

### DIFF
--- a/docs/docs/recoil-sync/api/RecoilURLSyncTransit.md
+++ b/docs/docs/recoil-sync/api/RecoilURLSyncTransit.md
@@ -14,6 +14,8 @@ function RecoilURLSyncTransit(props: {
 }): React.Node
 ```
 
+- **`handlers`** - Optional array of handlers for [Custom Classes](#custom-classes). It is important that this is a stable or memoized array instance. Otherwise you may miss URL changes as the listener is re-subscribed.
+
 ---
 
 Transit encoding is not as terse or readable as just using [JSON](/docs/recoil-sync/api/RecoilURLSyncJSON), however it can support `Map` and `Set` JavaScript containers as well as custom user classes.
@@ -21,8 +23,6 @@ Transit encoding is not as terse or readable as just using [JSON](/docs/recoil-s
 ## Custom Classes
 
 Handlers for custom user classes can be defined with the `handlers` prop:
-
-- **`handlers`** - Optional array of handlers for [Custom Classes](#custom-classes). It is important that this is a stable or memoized array instance. Otherwise you may miss URL changes as the listener is re-subscribed.
 
 ```tsx
 type TransitHandler<T, S> = {
@@ -39,7 +39,7 @@ type TransitHandler<T, S> = {
 class ViewState {
   active: boolean;
   pos: [number, number];
-  constructor(active, pos) {
+  constructor(active: boolean, pos: [number, number]) {
     this.active = active;
     this.pos = pos;
   }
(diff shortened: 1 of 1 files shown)


docs/docs/recoil-sync/api/RecoilURLSyncTransit.md
```diff
@@ -21,51 +21,55 @@ Transit encoding is not as terse or readable as just using [JSON](/docs/recoil-s
...
   active: boolean;
   pos: [number, number];
   constructor(active, pos) {
     this.active = active;
     this.pos = pos;
   }
-  ...
-};
-const viewStateChecker = custom(x => x instanceof ViewState ? x : null);
+  // ...
```
wd-fb, line 44: commented this out so prettier can recognize the syntax as valid and do it's thing

docs/docs/recoil-sync/api/RecoilURLSyncTransit.md
```diff
@@ -7,7 +7,7 @@ A component from the [Recoil Sync library](/docs/recoil-sync/introduction) to sy
 
 This is identical to the [`<RecoilURLSync>`](/docs/recoil-sync/api/RecoilURLSync) component except that it provides built-in [Transit encoding](https://github.com/cognitect/transit-js).
 
-```jsx
+```tsx
```
drarmstr, line 10: Did you notice a problem with `jsx` needing to switch to `tsx`?

drarmstr: It's probably a good idea to switch all the example/documentation syntax from Flow to TypeScript, though I do worry a tad about consistency...

docs/docs/recoil-sync/api/RecoilURLSyncTransit.md
```diff
@@ -21,51 +21,57 @@ Transit encoding is not as terse or readable as just using [JSON](/docs/recoil-s
...
-  write: T => S,
-  read: S => T,
+  tag: string;
+  class: Class<T>;
+  write: (T) => S;
+  read: (S) => T;
 };
 ```
 
+Note: ensure the value of the `handlers` prop is a stable constant.
```
drarmstr, line 34: Probably a good idea to be more forecful with this and include it directly in the main documentation above and not just here by the example.  Maybe something like:
```
  * **`handlers`** - Optional array of handlers for [Custom Classes](#custom-classes).  It is important that this is a stable or memoized array instance.  Otherwise you may miss URL changes as the listener is re-subscribed.

File: docs/docs/recoil-sync/api/RecoilURLSyncTransit.md (hunk 2 above)
Comment (wd-fb) [reply docs/docs/recoil-sync/api/RecoilURLSyncTransit.md:10]: ah yes. with JSX, prettier seemed to mangle the generic it doesnt understand (i'm editing the MDs within vscode if that matters):

```

const ViewState =
  atom <
  ViewState >
  {
    key: 'ViewState',
    default: new ViewState(true, [1, 2]),
    effects: [syncEffect({storeKey: 'transit-url', refine: viewStateChecker})],
  };
```

drarmstr: My VSCode is a mess editing the docs...  Check out how it actually looks with `/docs/yarn start`

wd-fb: Hmm what do you see?   Seems good on my local:

image

drarmstr: Oh, I just meant in general I can't really rely on VSCode for prettier or linting the docs, so I edit it there but rely on `/docs/yarn start` to see how it actually looks.

drarmstr (approved):

PR #2035, opened 2022-09-24
[recoil-sync] Expose updateItems in RecoilSync's Listen Interface

### LINKED ISSUE #2017: [recoil-sync] Support transactional updateItems() in RecoilSync
When syncing from an external store to recoil in the `listen` prop of RecoilSync, each call to `updateItem(k, v)` notifies components associated with each atom's key. However, since not all the state is synced yet (more synchronous `updateItem()` calls pending), some components will observe a "partially" synced state, which can cause errors.

In my case, this is making some components observe deleted state that would otherwise be consistent if all updates were applied before components get notified. 

Using updateAllKnownItems might work, but in my case, the number of atoms is vast and updates are often, so this will do a lot of unnecessary work (assuming updateAllKnownItems has transaction semantics).

luishdz1010:
Fixes #2017

### DIFF
--- a/packages/recoil-sync/RecoilSync.js
+++ b/packages/recoil-sync/RecoilSync.js
@@ -52,9 +52,11 @@ export type ReadItem = ItemKey =>
   | mixed;
 
 export type UpdateItem = <T>(ItemKey, DefaultValue | T) => void;
+export type UpdateItems = ItemSnapshot => void;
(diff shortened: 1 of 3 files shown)


FezVrasta:
Was this included in the 0.7.6 release?

drarmstr:
Included in `recoil-sync` 0.2.0 release.

PR #2046, opened 2022-09-30
improve API / naming for runtime env settings, expose through public recoil API

wd-fb:
Summary: followup on previous diff

Differential Revision: D39797419


brahn:
@drarmstr @wd-fb thanks for pushing this over the finish line!

PR #2089, opened 2022-11-06
fix: remove depricated `substr`

Gumichocopengin8:
- I replaced `substr` with `substring` since `substr` is now deprecated https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


Gumichocopengin8:
I signed CLA. Why is ci still failed?

PR #2091, opened 2022-11-06
feat: introduce `<RecoilURLSyncQueryString>`

Gumichocopengin8:
## Motivation

https://github.com/facebookexperimental/Recoil/discussions/2081#discussioncomment-4022664

> Is it possible to remove "" from query params when refine type is string? Currently it's like ?query="value", it would be nice to have ?query=value. That would be ideal to handle query params like [query-string](https://www.npmjs.com/package/query-string)
>> The quoting or other encoding is not part of the core <RecoilSync> or even the URL syncing but is rather part of the encoding layer on top such as <RecoilURLSyncJSON> `. The quotes are added as part of the JSON encoding of strings. I think it makes sense to add an additional encoding type for standard simple URL params, such as unquoted strings and vector entries. Please feel free to submit a PR for this!

## TODO
- [ ] Setup query-string configs
- [ ] Write test for them
- [x]  Throw error when param option is used in RecoilURLSyncQueryString
- [ ] Add types for Typescript
- Write doc
  - [ ]  No object support cuz query-string doesnt support it
  - [ ]  Mention no `param` option

Hello @drarmstr, I implemented the feature I asked before. I'm not sure if I implemented it in the right way, so I didn't write docs or full test cases because I'd like to hear your opinion first. 

First, I made a new component called `<RecoilURLSyncQueryString>` (temporary name). It is almost similar to `<RecoilURLSyncJSON>`, but it is serialized/deserialized by [query-string](https://www.npmjs.com/package/query-string) (I explored other libs and thought about not using libs at all, but `query-string` looked good.  If you know good  idea, I'm open). However, `query-string` has few restrictions. 

- Not recommended to use `object` in query parameters https://github.com/sindresorhus/query-string#nesting
  - So, I've decided not to support `object`
- `param` option is not applicable

### DIFF
--- a/packages/recoil-sync/RecoilSync_URL.js
+++ b/packages/recoil-sync/RecoilSync_URL.js
@@ -55,6 +55,7 @@ function parseURL(
   href: string,
   loc: LocationOption,
   deserialize: string => mixed,
+  isQueryString: boolean,
 ): ?ItemSnapshot {
   const url = new URL(href);
   switch (loc.part) {
@@ -71,6 +72,9 @@ function parseURL(
     case 'queryParams': {
       const searchParams = new URLSearchParams(url.search);
       const {param} = loc;
+      if (isQueryString) {
+        return wrapState(deserialize(url.search));
+      }
       if (param != null) {
         const stateStr = searchParams.get(param);
         return stateStr != null ? wrapState(deserialize(stateStr)) : new Map();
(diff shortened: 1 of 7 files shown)


packages/recoil-sync/RecoilSync_URL.js
```diff
@@ -55,6 +55,7 @@ function parseURL(
   href: string,
   loc: LocationOption,
   deserialize: string => mixed,
+  isQueryString: boolean,
```
drarmstr, line 58: We should really try to support this without leaking the implementation through the abstraction.  Seems like this is looking to avoid wrapping to/from the `ItemSnapshot` Map, can you handle that distinction in `RecoilURLSyncQueryString`?  Is the issue here because `query-string` is trying to handle the complete query parameter string while `queryParams` mode handles individual parameters?  There may be a concern because part of the intention of `queryParams` mode is to be able to co-exist with other query parameters that the URL may have which need to pass through.  (I'll update the unit tests to cover that as well for JSON). Given the need to support arrays via multiple entries, I see where the existing `queryParams` loop is problematic..   Trying to think of a cleaner solution...

drarmstr (changes requested):

packages/recoil-sync/RecoilSync_URL.js
drarmstr: Update the JSON tests as discussed in #2094

Gumichocopengin8: Thanks for the review. 
I also a bit concerned about this. I'll think of the better solution.
We can still use `query-string`, right?

drarmstr: I think it should be ok.. It uses the MIT license.  It does add some bloat, of course..

Trying to think if there's some solution leveraging the `pick()` method..

Gumichocopengin8: > We should really try to support this without leaking the implementation through the abstraction.

@drarmstr I tried some ways without touching the abstraction, but It seems difficult to implement this feature without editing the abstraction because `queryString.stringify()` generates entire query params, however, `encodeURL()` uses [searchParams.toString()](https://github.com/facebookexperimental/Recoil/blob/main/packages/recoil-sync/RecoilSync_URL.js#L121). These are not compatible. 
Ofc, if we use `replace()`, I think we can implement it, but I don't think it's a good idea.

> Trying to think if there's some solution leveraging the pick() method..

If I understand it correctly, [pick()](https://github.com/sindresorhus/query-string#pickurl-keys-options) wouldn't be helpful in this case. 

Do you have any good idea to implement this feature without editing the abstraction?

If we can't find a good way, I don't mind closing this PR.

PR #2116, opened 2022-12-04
Improve performance by compressing in parallel ✈️

sanjaiyan-dev:
Improving the performance by running the independent `async` functions in parallel ✈️.

_Sorry if I made any mistakes :(_


mondaychen:
That makes sense! Thanks!!

PR #2208, opened 2023-04-01
[Migration] Fix open source tests - remove extra parameter from ReactDOM.render()

drarmstr:
It appears that this commit https://github.com/facebookexperimental/Recoil/commit/b94eb51c2049d155984869b992e9ea8cf85d4101 (related to diff D44271426) broke the open-source tests by adding a string parameter to `ReactDOM.render()` from `ReactDOMLegacy_DEPRECATED`, while open source uses `react-dom`.  

I patched this using the `// @oss-only` and `// @fb-only` tags.  Though, if the source of truth migration is fully complete then that should no longer be necessary?


mondaychen:
oops, looks like this was merged along with #2209?

PR #2214, opened 2023-04-13
Add flag to skip circular dependency detection

RyanHoldren:
Summary:
I discovered a bug in `detectCircularDependencies()` where it may erroneously detect a circular dependency when `graphQLQueryEffect()` is used. It arises from the fact that we are using a global stack to keep track of which Atom/Selectors have been *visited*.

From the stack trace, I think I've figured out what is happening...
1) There are a bunch of `selectorGet()` calls to load dependencies.
2) This calls `detectCircularDependencies()` several times and a bunch of keys get added to the `dependencyStack`, as you would expect. So far, everything looks good.
3) Then, because it is a nested dependency, we load an Atom that has a `graphQLQueryEffect()`.
4) The `graphQLQueryEffect()` subscribes to a `RelayObservable`.The act of subscribing schedules a task using the `RelayFBScheduler`.
5) The schedule invokes `flushSyncCallbacks()` in ReactDOM.
6) Which calls `renderWithHooks()` in ReactDOM.
7) Which renders a component that calls `useRecoilValue()`.
8) We then end up back in `detectCircularDependencies()`, which is where we fail because the `dependencyStack` already includes the keys of all those previous Selectors and Atoms up to and including the Atom with the `graphQLQueryEffect()`.

This change adds a flag prop named `skipCircularDependencyDetection_DANGEROUS` to `<RecoilRoot>`, which is propagated down to the `Store`. If this flag is set to `true` then `detectCircularDependencies()` is bypassed.

This is obviously *less than ideal*. I think we can all agree that a fix for `detectCircularDependencies()` would be a much better solution. I am under the impression that, that will be a major refactor though, since it will require changing `dependencyStack` to be non-global and passing it all around. This flag will at least mitigate the bug until a more proper solution can be implemented. It should be very easy to back it out once that happens.

Differential Revision: D44563103

### DIFF
--- a/packages/recoil/core/Recoil_RecoilRoot.js
+++ b/packages/recoil/core/Recoil_RecoilRoot.js
@@ -65,6 +65,7 @@ type InternalProps = {
   initializeState?: MutableSnapshot => void,
   store_INTERNAL?: Store,
   children: React.Node,
+  skipCircularDependencyDetection_DANGEROUS?: boolean,
 };
 
 function notInAContext() {
@@ -365,6 +366,7 @@ function RecoilRoot_INTERNAL({
   initializeState,
   store_INTERNAL: storeProp, // For use with React "context bridging"
   children,
+  skipCircularDependencyDetection_DANGEROUS,
 }: InternalProps): React.Node {
   // prettier-ignore
   // @fb-only: useEffect(() => {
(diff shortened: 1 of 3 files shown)


itamark:
@RyanHoldren I think the failed check means you need to separate the ICSP and fbonly files out of the diff connected to this into a separate diff.

RyanHoldren:
I _think_ it's okay. Doesn't it just mean that this pull request can't be merged via GitHub but must be shipped via Phabricator?

> This Pull Request is synced with a Phabricator Diff that has internal-only changes. It is not safe to land directly on GitHub

PR #2228, opened 2023-04-28
[Docs] Update react docs link

bb8dd:
- Update links to old React docs to new React docs links


leonardovin:
Greeat we need this

PR #2244, opened 2023-06-06
RecoilRoot is also a Suspense boundary

kassens:
Summary:
Wraps the ReactRoot internally in a React Suspense boundary as Recoil uses `useEffect` in ways to do automatic batching that's not compatible with React 18 in case the Root suspends itself.

The reason for this is that with React 18, the `useEffect` callback is only called one the component is actually getting displayed. If the `useEffect `callback itself is what makes the component eventually un-suspend it causes an infinite suspense.

This isn't an ideal solution, but I'm not deep enough into Recoil to be able to propose a better solution.

Reviewed By: mrv1k

Differential Revision: D46405179


DouglasArmstrong-at:
@kassens - Should this also update the enabled GKs for OSS then: https://github.com/facebookexperimental/Recoil/blob/main/packages/shared/util/Recoil_RecoilEnv.js#L25

PR #2249, opened 2023-06-14
Corrected the wrong message while copying failed for README.md file

tajnur007:
Here, if the build script failed to copy the **README.md** file then it is showing **Failed to copy README-recoil.md**. But there is no file named as README-recoil.md into project's root directory, the message should be **Failed to copy README.md**. Rest of the changes into this file done by prettier.

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -73,8 +73,8 @@ async function buildAll() {
   console.log('Building all packages...\n');
   await Promise.all(
     Object.entries(PACKAGES).map(([target, config]) =>
-      buildPackage(target, config)
-    )
+      buildPackage(target, config),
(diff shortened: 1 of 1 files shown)


scripts/build.mjs
```diff
@@ -104,7 +104,9 @@ async function buildPackage(target, config) {
     `${projectRootDir}/README${target === 'recoil' ? '' : '-' + target}.md`,
     `${BUILD_TARGET}/${target}/README.md`,
     fs.constants.COPYFILE_FICLONE,
-    createErrorHandler(`Failed to copy README-${target}.md`),
```
tajnur007, line 107: This PR is specifically related to this line only.

PR #2252, opened 2023-06-26
reap test recoil_mutable_source

kassens:
Summary:
Reaps the GK to use `useMutableSource` in Recoil. While the GK is at 100%, it's never hit as we hit the check for `recoil_sync_external_store` before and that's also 100%.

`useMutableSource` is deprecated in React and this is one of the last callers.

Differential Revision: D47023855


noahlemen:
should we fully ship the `recoil_sync_external_store` case as well considering its also at 100%?

kassens:
Wanted to take this one step at a time, but this one seemed like a save bet to me as useMutableSource is deprecated in favor of syncExternalStore.

PR #2275, opened 2023-08-20
only set isSSR true if !isReactNative

### LINKED ISSUE #2274: Infinite suspense with async selector on React Native
I believe this is a regression introduced in 0.7.7 specifically https://github.com/facebookexperimental/Recoil/commit/cff501505dc0491089627622fb1c677dba3c21e1

The problem seems to be that `isSSR` is evaluating to true in the react native environment, thus executing the this resolver delete statement:
https://github.com/facebookexperimental/Recoil/blob/dba6c40b8f2305904abdcacd1f191bc9153812d7/packages/recoil/hooks/Recoil_Hooks.js#L69-L75

Not sure if that's a new thing - I was just trying this out with react-native 0.72.3 so that may be due to a recent change in react native.

PR coming in a moment

tplorts:
fixes #2274 

I couldn't find any tests around these environment conditions.  Seems like something that would be difficult to make a good test for but let me know if there are tests I could update.


coolersham:
I have tested this PR in my project, it solves the linked issue and therefore should be merged and released ASAP.

wojtekmaj:
I have also tested this patch and I confirm it has resolved the issue I was having reported under #2274. I hope this gets merged soon.

PR #2295, opened 2023-12-08
fix: atomRegistry deletion overlaps in the cleanup stage

### LINKED ISSUE #1994: [BUG] Snapshot has already been released.
From [#1891](https://github.com/facebookexperimental/Recoil/pull/1891),
this issue is not fixed at all in dev mode (firefox, safari browser).
when debug console is opened or after page refresh, it keeps throwing the error.
Could you possibly add examples of how to use urlSyncEffect properly?
There's so little examples on the official web site and I have no idea what has been done wrong.

cjvnjde:
I found an issue particularly noticeable in strict mode: initializing two effects followed by a cleanup operation erroneously removes cached data for both effects. 
This issue becomes particularly apparent in environments combining strict mode with Next.js. To illustrate, consider the following sequence of events (real sequence on my project):
```
set key:  testState       1
delete key:  testState    0
set key:  testState       1
set key:  testState       2
delete key:  testState    0 // Issue: should be 1 because only one deletion occurred
```
The root cause of this issue is the reuse of the same key for multiple 'set' operations, leading to incorrect 'delete' behavior. A single cleanup call affects all instances with the same key rather than the specific instance intended.

To resolve this, we need to use a distinct key for each 'set' operation, and the corresponding 'delete' operation targets this specific key. This approach ensures accurate tracking and removal of individual effects.

Additionally, this fix also fixes [this](https://github.com/facebookexperimental/Recoil/issues/1994) issue. The reason for that problem was the deletion of the "second key" that had been deleted with the previous cleaning.

### DIFF
--- a/packages/recoil-sync/RecoilSync_URL.js
+++ b/packages/recoil-sync/RecoilSync_URL.js
@@ -39,7 +39,10 @@ type AtomRegistration = {
   itemKeys: Set<ItemKey>,
 };
 
-const registries: Map<StoreKey, Map<NodeKey, AtomRegistration>> = new Map();
+const registries: Map<
+  StoreKey,
+  Map<{nodeKey: NodeKey}, AtomRegistration>,
+> = new Map();
 
 const itemStateChecker = writableDict(mixed());
 const refineState = assertion(itemStateChecker);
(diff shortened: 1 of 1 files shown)


FezVrasta:
I'm experiencing https://github.com/facebookexperimental/Recoil/issues/1994 just using `syncEffect` while I'm trying to upgrade from React 17 to 18, on Chrome.

PR #2315, opened 2024-04-24
feat: removing the word 'experimental' from the README.

hammygoonan:
Given Recoil is 4 years old now, I think it's safe to say it is no longer experimental.


clockk:
Although four years have passed since its release, the major version v1.x has not yet been released.
What do you think about this?

